PR #37, opened 2026-03-22
add clang format and clang tidy checks

BrainTeazer:
Having a consistent and modern codebase would help with project maintenance. This PR:

- Adds clang-format file/configuration
- Adds clang-tidy file/configuration
- Adds a github workflow to check for linter warnings/errors, only on files that a PR touches
- Fixes an error in wayland_subsurface.cpp when wayland-protocols<1.47 that creates errors for the workflow

Currently, it runs on an ubuntu image, so the clang-tidy checks may not work for Windows/MacOS changes. However, clang-format should work regardless.

Open to suggestions for the preferred clang-format config and clang-tidy checks. 

**IMPORTANT**:
A new ruleset would need to be created for the project to include this workflow in the PR checks. 
The ruleset would need to target only the default branch (main) and would need to enable '**Require status checks to pass**' and then add the cpp-linter status check.


.github/workflows/cpp-linter.yml
```diff
@@ -0,0 +1,49 @@
...
+            wayland-protocols libplacebo-dev libsystemd-dev \
+            plasma-wayland-protocols meson ffmpeg \
+            libavcodec-dev libavformat-dev libavfilter-dev \
+            libass-dev libwayland-dev
+
+      - name: Download and generate dependencies, and build
+        run: | 
+          git submodule update --init --recursive
+          python3 dev/download_cef.py
+          cmake -B build -G Ninja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
```
zjeffer, line 33:
```suggestion
          cmake -B build -G Ninja
```
`-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` is unnecessary as it's already defined in CMakeLists.txt

.github/workflows/cpp-linter.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: libsdl-org/setup-sdl@main
+        id: sdl
+        with:
+          install-linux-dependencies: true
+          version: 3-latest
+          version-sdl-image: 3-latest
```
zjeffer, line 17: sdl has been removed from the repo

andrewrabert:
Not sure I want to add this at this time. Maybe once the code base settles into stability.

zjeffer:
I feel like the code needs at least a clang-format file. Right now none of the code is formatted which is really annoying because I often try to format my changes but then it changes the formatting for the entire file, which I obviously don't want to put in a PR. 
Same thing with the javascript code, which is currently not formatted at all.

BrainTeazer:
I agree with adding it sooner rather than later. As the code base matures, the more changes will need to be made to adhere to clang-tidy warnings. Also, it would also be great to have a consistent formatter right now. 

The Github Action only checks for changed files, but formatting the entire codebase can be done with a one line command (something like below). So it should be painless.

```
find foo/bar/ -iname '*.h' -o -iname '*.cpp' | xargs clang-format -i
```

Maybe for now it could be a draft PR rather than a closed PR?

Issue #20, opened 2026-03-15
Wrong clipboard content type pasted


andrewrabert:
The wrong clipboard content is pasted into CEF when the clipboard has, for example, html and plain text content. 

```
$ wl-paste --list-types
text/html
text/_moz_htmlcontext
text/_moz_htmlinfo
text/plain;charset=utf-8
UTF8_STRING
COMPOUND_TEXT
TEXT
text/plain
STRING
text/plain;charset=utf-8
text/plain
text/x-moz-url-priv
SAVE_TARGETS

$ wl-paste --type text/html
<meta http-equiv="content-type" content="text/html;
...
````

## Steps to reproduce
1. Go to https://github.com/jellyfin-labs/jellyfin-desktop-cef/issues/20
2. Select the text as shown highlighted in the image. <img width="575" height="169" alt="Image" src="https://github.com/user-attachments/assets/b2dde42b-1854-44d6-a363-d849f3cc7d4b" />
3. Ctrl+c
4. Open Jellyfin Desktop CEF
5. Go to an instance's search page
6. Ctrl+v

Notice how it pasted html content.
<img width="695" height="340" alt="Image" src="https://github.com/user-attachments/assets/3622309a-6f4b-45b0-8cf0-94908a743dd9" />

The below image is what happens when pasting into the search box within Firefox on Linux. (ignore the text being cutoff as that's unrelated to anything in desktop client.)
<img width="634" height="294" alt="Image" src="https://github.com/user-attachments/assets/0ade76b7-cc7c-440e-b90a-d48d0464c8c2" />

andrewrabert:
Should be fixed by https://github.com/jellyfin/jellyfin-desktop/pull/110

PR #39, opened 2026-03-22
fix(macOS): fix macOS paste not working

### LINKED ISSUE #28: Macos paste does not work
**Describe the bug**
Cmd-v seems to do nothing when I have text in a buffer

**Steps to reproduce**
Try pasting on the server address or search bar, nothing happens

**System information**
 - Installation Method: dmg download from readme
 - OS: MacOS 26.2 M2

BrainTeazer:
macOS paste does not work, because the mimetype of plaintext is "text/plain;charset=utf-8". Now, SDL_GetClipboardText is used to get the clipboard text if it is text, otherwise the allowed mime types are checked using SDL_GetClipboardData.

Fixes #28

### DIFF
--- a/src/input/browser_layer.h
+++ b/src/input/browser_layer.h
@@ -82,19 +82,27 @@ public:
                     bool shift = mods & EVENTFLAG_SHIFT_DOWN;
                     switch (event.key.key) {
                         case SDLK_V: {
-                            static const char* mimeTypes[] = {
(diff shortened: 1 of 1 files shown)


BrainTeazer:
Seems to also fix #20 on macOS.

andrewrabert:
thanks!

Issue #12, opened 2026-03-09
Blackscreen in Fullscreen Mode


Niwu34:
Asa I swich to fullscreen on Win 11 the playback and Gui turns black, music keeps playing.
Happens with non HDR and SDR vids, turning on/off HDR in win settings doesn't change anything

Vid below for clarity to showcase: (Switch to fullscreen and back)

https://github.com/user-attachments/assets/3d146211-8c1e-443f-8706-fecef3d221ed

andrewrabert:
This _might_ be fixed, though I wasn't able to replicate it at the time you initially submitted the issue. 

Since then, there was a hefty refactor of the Windows rendering pipeline starting with https://github.com/jellyfin-labs/jellyfin-desktop-cef/commit/1fe9fb57d5ff7972615bbdc354d80433768bb96e

Try the latest build and left me know if it's still showing black in fullscreen.

xBiei:
This is actually happening to me on wayland, so its not a windows-only issue.

xBiei:
after some digging, it looks like its always using pass-through whether the content is HDR or not

```
[Platform] Using PASS_THROUGH with R16G16B16A16_UNORM (format 91)
[Platform] Swapchain: 3428x1428 format=91 colorSpace=1000104013 HDR=yes
```

and even if the content is HDR, it always tells the CM to switch to HDR but keeps showing a black screen.

xBiei:
Fixed after switching to libplacebo swapchain.

MarcoCoreDuo:
Sadly fullscreen is still broken for me on Fedora with nvidia and wayland. The UI is still visible though. It's only the mpv background layer that doesn't work

andrewrabert:
@MarcoCoreDuo what do you use?
- Fedora version
- GPU model
- GPU driver

MarcoCoreDuo:
> [@MarcoCoreDuo](https://github.com/MarcoCoreDuo) what do you use?
> 
> * Fedora version
> * GPU model
> * GPU driver

- Fedora Linux 43 (Workstation Edition)
- NVIDIA GeForce RTX 4080
- 595.45.04

MarcoCoreDuo:
New findings:
- When I enable HDR for the monitor, fullscreen works there.
- When only SDR is enabled, fullscreen only works if the mouse pointer is not on the monitor where fullscreen is active.

Dangonizer:
Same thing on Plasma Wayland with AMD
[...]

PR #57, opened 2026-03-28
fix black window on wayland

Renari:


Renari:
This resolves #12 but only on Wayland, this does not solve the same issue happening on Windows.

andrewrabert:
Thanks!

joshuaboniface:
Confirmed fixes the bug (Debian 13 w/GNOME + Wayland)!

PR #63, opened 2026-03-29
feat: add --tv flag for fullscreen TV mode

n1coh:
## Summary
- Adds a `--tv` CLI flag that starts the window in fullscreen and enables the jellyfin-web TV layout
- Sets `fullscreen: true` in the NativeShell config so jellyfin-web renders the TV UI
- Matches the `--tv` flag from the Qt-based v2.0.0

## Test plan
- [ ] Run `jellyfin-desktop --tv` and verify it starts in fullscreen with TV layout
- [ ] Run `jellyfin-desktop` without flag and verify normal windowed behavior is unchanged
- [ ] Verify `jellyfin-desktop --help` shows the new `--tv` option


andrewrabert:
Will re-evaluate this in the future.

Issue #30, opened 2026-03-18
Side scrolling


bilbofroggins:
**Describe the feature**
Side scrolling would be nice on the browse page. I usually have to horizontal scroll a lot and I don't want to have to click the left right buttons

**Did Jellyfin Media Player / the Qt desktop client have this feature?**
Yes

andrewrabert:
Working as of https://github.com/jellyfin/jellyfin-desktop/pull/110

PR #64, opened 2026-03-29
Implement native wheel event for macOS CEF

### LINKED ISSUE #29: Scrolling is jittery
**Describe the bug**
While scrolling on the browse page, there are noticeable abrupt jumps. Smooth scrolling is default on mac

**Steps to reproduce**
Scroll a small amount with something like a trackpad, the page will jump down instead of smooth scrolling

**System information**
 - Installation Method: dmg from readme
 - OS: macOS 26.2 M2

gnattu:
For macOS, the old behavior is handling wheel events in the SDL layer and adding a 10x scale multiplier to match the scroll speed the OS would normally have. While this approach works, it results in a less smooth scrolling experience because we discard all intermediate states between scrolling events. Consequently, the scrolling “framerate” is significantly lower than it should be, as the scroll position between two events are huge enough to look like a jump visually.

To fix this, we need to capture native wheel events using macOS’s native method and forward them to CEF. This will allow CEF to handle the events in a manner similar to a typical Chromium-based browser.

Fixes #29
This would also address #30 for macOS, as now trackpad x axis scrolling is also correctly handled.

### DIFF
--- a/src/cef/cef_client.cpp
+++ b/src/cef/cef_client.cpp
@@ -6,6 +6,7 @@
 #include "include/cef_parser.h"
 #include <SDL3/SDL.h>
 #include "logging.h"
+#include <cmath>
 #include <mutex>
 #if !defined(__APPLE__) && !defined(_WIN32)
 #include <unistd.h>  // For dup()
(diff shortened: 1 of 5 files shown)


andrewrabert:
Thanks!

andrewrabert:
Reverted as it broke scroll wheel on my mouse. (Excruciating slow)

Issue #90, opened 2026-04-02
Manual MPV Configuration section is missing under Client Settings


bjevers:
**Describe the bug**
Manual MPV Configuration no longer appears under Client Settings

**Steps to reproduce**
1. Click menu icon
2. Click Client Settings and scroll, it's missing

**System information**
 - Installation Method: April 1st action run
 - OS: Windows 11 25H2
 - GPU: Nvidia RTX 4080 Super

**Additional context**
This option is needed for SVP 4 Pro to function.  Also for using custom shaders.

PR #96, opened 2026-04-03
Custom mpv settings

rodis120:
Added a way to pass custom options to MPV player.
The code was inspired by the implementation from [jellyfin-desktop-gt](https://github.com/jellyfin-archive/jellyfin-desktop-qt).

Client config:
image

This PR solves issues: #90 #129

### DIFF
--- a/src/browser/web_browser.cpp
+++ b/src/browser/web_browser.cpp
@@ -11,6 +11,7 @@
 #include "../titlebar_color.h"
 #include "../input/dispatch.h"
 #include "../cjson/cJSON.h"
+#include "../paths/paths.h"
 
 extern void update_idle_inhibit();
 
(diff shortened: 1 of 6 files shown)


rodis120:
@andrewrabert I updated this PR, so it is now compatible with your rewrite https://github.com/jellyfin/jellyfin-desktop/pull/110

andrewrabert:
I'm considering a different approach: a button to open the config file in a text editor and a button to open the mpv config dir.

The app already reads standard mpv config from ~/.config/jellyfin-desktop/mpv/config

RekiDunois:
In macos I put mpv.conf into ~/.config/jellyfin-desktop/mpv/config but it does not work

rodis120:
@andrewrabert That seems like a more elegant and cleaner solution, also it allows for defining custom keybinds  in `~/.config/jellyfin-desktop/mpv/input.conf`. I will try to implement that buttons.

rodis120:
@andrewrabert I rewritten my PR so now it only adds a button that opens mpv config directory. I tested it on linux and windows. I didn't test it on macos as I don't have access to any apple device, but I think it should work on it as well.

obraz

andrewrabert:
Awesome, thanks!

PR #99, opened 2026-04-03
fix(macOS): improve native wheel handling for mice [v2]

### LINKED ISSUE #29: Scrolling is jittery

gnattu:
Compared to v1, we no longer treat all macOS scroll input as “precise” pixel scroll because not all mice, especially third-party mice with traditional wheels, do not have pixel-perfect scroll events. V1 mistakenly treated all scroll events as pixel-precise, which broke scrolling on such mice and made it painfully slow.

For precise devices like Apple branded mice and trackpads, the behavior stays close to v1: native deltas are forwarded with `EVENTFLAG_PRECISION_SCROLLING_DELTA`. 

For non-precise wheel mice, we now use AppKit’s accelerated `event.deltaX/Y` instead of `scrollingDeltaX/Y`, this allows the CEF experience more "macOS native" mouse wheel handling, where the wheel scrolls faster, the sensitivity goes higher.

We also tried to replicate the smooth scrolling for mice without pixel-precise scroll by turning discrete wheel inputs into short pending scroll impulses. These impulses are then drained across multiple frames, which helps reduce jumpiness and creates a more Chrome-like smooth wheel feel. But chrome has much more work beyond this so we are not as smooth as real Chrome in this case, but it is much better than current brute-force way, and implementing overly complicated solution for scrolling is clearly out of scope of this project.

v1 changes, for reference
-----

For macOS, the old behavior is handling wheel events in the SDL layer and adding a 10x scale multiplier to match the scroll speed the OS would normally have. While this approach works, it results in a less smooth scrolling experience because we discard all intermediate states between scrolling events. Consequently, the scrolling “framerate” is significantly lower than it should be, as the scroll position between two events are huge enough to look like a jump visually.

To fix this, we need to capture native wheel events using macOS’s native method and forward them to CEF. This will allow CEF to handle the events in a manner similar to a typical Chromium-based browser.

Fixes https://github.com/jellyfin/jellyfin-desktop/issues/29
This would also address https://github.com/jellyfin/jellyfin-desktop/issues/30 for macOS, as now trackpad x axis scrolling is also correctly handled.

### DIFF
--- a/src/cef/cef_client.cpp
+++ b/src/cef/cef_client.cpp
@@ -6,6 +6,7 @@
 #include "include/cef_parser.h"
 #include <SDL3/SDL.h>
 #include "logging.h"
+#include <cmath>
 #include <mutex>
 #if !defined(__APPLE__) && !defined(_WIN32)
 #include <unistd.h>  // For dup()
@@ -718,6 +719,8 @@ void Client::sendMouseWheel(int x, int y, float deltaX, float deltaY, int modifi
     scroll_x_ = x;
     scroll_y_ = y;
     scroll_mods_ = modifiers;
+    scroll_precise_ = true;
+    scroll_animated_ = false;
     accum_scroll_x_ += deltaX * 10.0f;
     accum_scroll_y_ += deltaY * 10.0f;
     has_pending_scroll_ = true;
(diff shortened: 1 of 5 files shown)


andrewrabert:
No longer SDL

PR #107, opened 2026-04-04
fix pathlib handling on posix systems

jonah-saltzman:
`is_junction` is not available on PosixPath objects in python, causing a `AttributeError: 'PosixPath' object has no attribute 'is_junction'` when running `dev/download_cef.py` script on MacOS (and presumably linux as well).

Add a platform check before calling is_junction on the path objects in this dev script.


andrewrabert:
I haven't seen an error about this on any platform. 

What OS/distro/Python version did you see it on?

jonah-saltzman:
On Python 3.9.25 (macos 26.3.1)

Switching to 3.13 did fix it actually so maybe this isn't necessary.

PR #108, opened 2026-04-05
reveal macos menu bar when moving mouse to top of screen

jonah-saltzman:
Currently when you enter fullscreen from Jellyfin client, the top menu bar is never visible even if moving the mouse to the top of the screen. 

This sets the `SDL_HINT_VIDEO_MAC_FULLSCREEN_MENU_VISIBILITY` hint, allowing us to reveal the top menu bar when mousing to the top of the screen:

Screenshot 2026-04-04 at 8 28 29 PM


andrewrabert:
Fixed in https://github.com/jellyfin/jellyfin-desktop/pull/110

jonah-saltzman:
Ah nice. Thanks for your hard work!

PR #119, opened 2026-04-07
Add startup fullscreen CLI flag

thankarezos:
## Description
This PR adds a new --fullscreen command-line flag to start Jellyfin Desktop in fullscreen mode.

## Why
In minimal and HTPC environments, reducing startup interaction is important. The previous Qt client effectively supported this workflow, and users migrating from it expect a direct fullscreen startup option.

## Behavior
- New opt-in flag: --fullscreen
- Default startup behavior is unchanged when the flag is not provided

## Testing
- Launch with --fullscreen: app starts in fullscreen
- Launch without --fullscreen: app starts with existing saved/maximized behavior
- --help shows the new flag

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -358,7 +358,7 @@ int main(int argc, char* argv[]) {
                        "  --audio-passthrough <codecs>  Enable audio passthrough (e.g. ac3,dts-hd,eac3,truehd)\n"
                        "  --audio-exclusive       Use exclusive audio output mode\n"
                        "  --audio-channels <layout>  Set audio channel layout (e.g. stereo, 5.1, 7.1)\n"
-                      "  --fullscreen            Start in fullscreen mode\n"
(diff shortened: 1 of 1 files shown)


thankarezos:
@andrewrabert Other that resolving the conflicts maybe this needs to be --kiosk (instead of --fullscreen) so it doesn't allow the player to minimize (and even better hide the full-screen button) this is much more complex on keeping the PR simple, so I would like some feedback about it on how you think something like that would plug on the current implementation (or if it even makes sense at all), so I might try to materialize the appropriate change.  But --kiosk mode would be very appreciated by anyone who uses Jellyfin on HTPCs where you want it to always start with full-screen and keep it full-screen. QT version's build in  --fullscreen flag could do that (maybe not perfectly cause there was still the full-screen button on the player even though it did nothing).

thankarezos:
Replaced by #169

PR #120, opened 2026-04-07
Linux: prefer X11 and fall back to Wayland when DISPLAY is unavailable

thankarezos:
## Description
This PR updates Linux CEF backend selection for `ozone-platform`.

## Why
Some environments do not have X11 available (`DISPLAY` is unset). In those cases, forcing X11 can cause startup failures. This change keeps X11 as the preferred backend while adding an automatic Wayland fallback.

## Behavior
- On Linux, if `DISPLAY` is available:
  - use `ozone-platform=x11`
- On Linux, if `DISPLAY` is unavailable:
  - use `ozone-platform=wayland` (fallback)
- Logs were added to make backend selection explicit:
  - `Using X11 backend (ozone-platform=x11)`
  - `DISPLAY unavailable; falling back to Wayland (ozone-platform=wayland)`

## Scope
- No new CLI flags.
- No behavior change on macOS or Windows.

## Testing
- Built and launched on Linux.
- Verified with `DISPLAY` set: selects X11.
- Verified without `DISPLAY`: falls back to Wayland.
- Verified logs reflect selected backend path.

### DIFF
--- a/src/cef/cef_app.cpp
+++ b/src/cef/cef_app.cpp
@@ -7,6 +7,7 @@
 #include "include/cef_frame.h"
 #include <cmath>
 #include <cstring>
+#include <cstdlib>
 #include "logging.h"
 
 void App::OnBeforeCommandLineProcessing(const CefString& process_type,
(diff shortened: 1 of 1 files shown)


andrewrabert:
To be handled in https://github.com/jellyfin/jellyfin-desktop/pull/154

I ended up figuring out how to get scaling working properly with the wayland ozone backend.

Issue #18, opened 2026-03-15
Mouse cursor autohide broken on Windows


andrewrabert:

g2kokaro:
As a workaround, the cursor hides if you wait a moment for the UI elements to auto-hide, then right click the mouse. After doing so, moving the mouse doesn't bring back the cursor or UI until you left click.

andrewrabert:
Fixed in https://github.com/jellyfin/jellyfin-desktop/pull/110

PR #121, opened 2026-04-07
Fix cursor auto-hide by syncing SDL cursor visibility with jellyfin-web mouseIdle

Mathis0d:
**This adapts the old Qt cursor fix from [jellyfin-desktop-qt commit 8009e81](https://github.com/jellyfin-archive/jellyfin-desktop-qt/commit/8009e817a604d924bed583785ab36b8ca571a0f5) to the current CEF/SDL client.**

**Fixes this issue: #18**

Instead of relying on `body.mouseIdle` forcing `cursor: none`, then waiting for CEF to emit `CT_NONE`, the web layer now forwards cursor visibility directly to the native side. The SDL layer keeps cursor visibility separate from cursor shape, so the previous cursor is restored immediately when activity resumes.

## Why

The current implementation can intermittently fail to hide the cursor during video playback because it depends on CEF cursor-change events being emitted when only the `mouseIdle` class changes.

## Changes

- add a native `setCursorVisibility` bridge
- observe `body.mouseIdle` directly in `native-shim.js`
- stop using injected `cursor: none` as the auto-hide trigger
- keep SDL cursor visibility separate from cursor shape restoration

## Testing

https://github.com/user-attachments/assets/f257654e-7fdf-46ed-9c65-0eb7716ab7f4


andrewrabert:
Should be fixed as of https://github.com/jellyfin/jellyfin-desktop/pull/110

PR #154, opened 2026-04-13
Add X11 platform support

### LINKED ISSUE #140: X11 Support
**Describe the feature**
Add X11 support. Currently only Wayland is supported

**Context**
The refactor https://github.com/jellyfin/jellyfin-desktop/pull/110 changed Jellyfin Desktop to overlay transparent CEF layers on top of mpv-created window (similar to standalone mpv player). This eliminated a complex libmpv vulkan fork that Jellyfin Desktop was previously using. The previous approach was becoming more complex to maintain - especially with regards to color management. There were also occasional stutters and crashes stemming from the libmpv fork.

Switching to the https://github.com/jellyfin/jellyfin-desktop/pull/110 approach eliminated all of that by reusing the same code that mpv standalone uses. Wayland, macOS, and Windows no longer need to care about things like HDR detect - mpv already does that and does it well.

For Linux specificly, the refactor resulted in exclusive reliance upon Wayland protocols for compositing, CEF rendering, and input - as they solved their respective problems _best_. Introducing X11 support would need a different, separate approach added to Jellyfin Desktop.

andrewrabert:
Adds X11 support.

Resolves https://github.com/jellyfin/jellyfin-desktop/issues/140

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,16 +116,29 @@ elseif(WIN32)
         windowsapp
     )
 else()
-    # Linux: native Wayland (mpv owns window, CEF on subsurfaces via dmabuf)
+    # Linux: Wayland + X11 backends (runtime selection)
     find_package(PkgConfig REQUIRED)
+
+    # Shared Linux deps
+    pkg_check_modules(SYSTEMD REQUIRED libsystemd)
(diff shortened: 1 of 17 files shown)


enter-a-random-username:
Remove the Wayland warning in README

PR #156, opened 2026-04-13
Fix server connect spinner hanging on trailing-slash URLs

TheArchitect0880:
Fixes a connect hang where the loading spinner never ends after entering a server URL with a trailing slash. This normalizes the URL by removing trailing slashes before matching connectivity results.


andrewrabert:
Thanks. 

Decided on a different approach and will be handling in https://github.com/jellyfin/jellyfin-desktop/pull/188 (and now there are tests for this :) )

PR #164, opened 2026-04-14
Upstream url for CEF

enter-a-random-username:
Replace mirror URL with upstream according to CEF webpage

e.g. referenced here
https://chromiumembedded.github.io/cef/contributing_with_git.html


andrewrabert:
ty!

PR #169, opened 2026-04-14
Add kiosk startup CLI flag and enforce fullscreen lock

thankarezos:
Description  
This PR adds a new startup flag, --kiosk, to launch Jellyfin Desktop directly in kiosk mode and keep fullscreen locked during playback and UI transitions.

Why  
In HTPC environments, startup friction should be minimal and accidental fullscreen exits should be prevented. Users need a mode that starts fullscreen immediately and stays fullscreen even when UI events, hotkeys, or renderer callbacks try to toggle window state.

Behavior  
- New opt-in flag: --kiosk  
- App starts in fullscreen when --kiosk is provided  
- Default startup behavior is unchanged when the flag is not provided  
- Kiosk mode is propagated to CEF child processes and exposed to the renderer as window.kioskMode  
- Fullscreen exit is blocked in kiosk mode for:
  - Keyboard toggles (F, F11)
  - IPC fullscreen toggle requests
  - CEF fullscreen false callbacks
  - OSD-driven auto-exit transitions
- Web fullscreen button is hidden in kiosk mode

Testing  
- Launch with --kiosk: app starts in fullscreen  
- While in kiosk mode:
  - Press F/F11: fullscreen does not exit
  - Trigger UI fullscreen toggle: fullscreen does not exit
  - Trigger OSD open/close cycle: fullscreen does not exit
  - Stop playback: fullscreen remains locked
- Launch without --kiosk: existing non-kiosk behavior remains intact
- Help output shows the new --kiosk flag

Future Work  
- Add optional ESC handling for kiosk mode on the home screen.  
- On ESC, show a confirmation dialog: “Exit Jellyfin Desktop?” with Cancel and Exit actions.  
- Guardrails:
  - Only active in kiosk mode.
  - Only on home/browse screens (not during playback).


andrewrabert:
Not a fan of propagating a special "kiosk" mode throughout the codebase versus a combination of smaller behaviors (allowed to exit fullscreen).

thankarezos:
@andrewrabert Ok got it. So if i split this into "--fullscreen" and "--allow-exit-fullscreen" would be ok by you? Is the Idea of blocking exit fullscreen and hidding the button ok as a behaviour?

Issue #163, opened 2026-04-14
Bug: volume doesn't save across sessions


cloudbells:
**Describe the bug**
Volume level doesn't save when closing the window. I've tried both exiting while unpaused, paused, and going back to home screen and closing it there. Seems to be not saving regardless of how it's closed.

**Steps to reproduce**
1. Start playback.
2. Lower volume.
3. Exit jellyfin-desktop.
4. Restart.
5. Open any show/movie and volume is back at 100%.

**System information**
 - Installation Method: AUR
 - OS: Arch Linux
 - GPU: e.g. AMD RX 7900 XTX
 - **X11** or **Wayland**? Wayland (Hyprland)

andrewrabert:
Fixed in https://github.com/jellyfin/jellyfin-desktop/pull/176

PR #176, opened 2026-04-15
Fix persistent volume and missing audio in music player

rodis120:
This PR fixes:
- Persistent volume across sessions
- Audio disappearing after changing volume in music player

Related issues: #163


andrewrabert:
Thanks!

PR #188, opened 2026-04-17
Rework server connect flow

### LINKED ISSUE #62: Client hangs waiting on server response.
**Describe the bug**
I changed my jellyfin server's address. The jellyfin-desktop application will hang if the server address is changed. There is no way to get back to the page that requests the server address. Users must change the config file manually. For linux it just hangs. For windows it hangs and memory leaks very quickly making the system unresponsive.

**Steps to reproduce**
1. Assuming a valid server and client connection; close the client, stop the server. Alter the address / url for the server. Start server, navigate there via the browser to double check the new url.
2. Open the client. It hangs. You cannot get to the "Server Address" page to correct the value.
3. Close the client.
4. Manually remove / null the "sections: main: userWebClient" value in linux's user config (~/.local/share/jellyfin-desktop/profiles/###/jellyfin-desktop.conf) or the corresponding setting in window's client user config. 
5. Start client, see "Server Address" page. 

**System information**
 - Installation Method: home manager
 - OS: nixos stable channel, https://search.nixos.org/packages?channel=25.11&query=jellyfin&show=jellyfin-desktop (its version 2.0.0)
 - GPU: none
 - (Linux Only) x11 xfce

**Additional context**
Stopping the server and then starting the client will cause the same issue. Even if you start the server after the client is in this hanging state it can't recover. 

An alternative method of reproduction could probably be to change the "userWebClient" in the client instead of the server's url.

andrewrabert:
- Extract URL handling into a `jellyfin_api` module with unit tests (normalization, `/web` stripping, `/System/Info/Public` response validation). Overlay probe uses HEAD + redirect-follow and is fully cancellable.
- Make Escape during auto-connect actually cancel: cancellable min-spinner, tear down the main-browser pre-load on cancel via a `CefLayer` close/recreate state machine, split `loadServer` into `navigateMain` + `dismissOverlay` so the pre-fade wait lives in JS.
- Route CEF renderer `console.*` into Quill as `[JS]` with a backend-thread newline-strip sink; collapse `LOG_JS_MAIN`/`LOG_JS_OVERLAY` into a single `LOG_JS`.
- `justfile` for the dev loop (`build`, `test`, `run`, `deps`, `clean`); vendored doctest; `BUILD_TESTING` off by default.

Fixed https://github.com/jellyfin/jellyfin-desktop/issues/62


s1gtrap:
I'm not sure if this actually fixes #62 or handles it like #156 would. What happens on my end on latest main (8a63828) is it'll read `serverUrl` from `~/.config/jellyfin-desktop/settings.json`, try to connect to it, and if it succeeds, for some unknown reason, it'll immediately overwrite it to an invalid value.

It will work as intended the first time, provided you manually edit `~/.config/jellyfin-desktop/settings.json` and fix `serverUrl`, but the next time you try to run it, it'll fail to connect because it overwrote your changes on the first successful run.

I'm not sure what the underlying issue is, but if you examine the logs, immediately after loading `~/.config/jellyfin-desktop/settings.json`, it prints the following to stdout:

```
[JS] Stored JSON credentials: {"Servers":[{"DateLastAccessed":1778671650727,"LastConnectionMode":2,"ManualAddress":"https://somedomain.xyz/jellyfin","manualAddressOnly":true,"Name":"Jellyfin","Id":"a22c288ff34d5119d30afdcb3d9b1a4a","LocalAddress":"http://10.42.4.28:8096/jellyfin","AccessToken":"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx","UserId":"9412e97c6a09c5f6b8fec075019bcea1"}]} (https://somedomain.xyz/jellyfin/web/node_modules.jellyfin-apiclient.bundle.js?bd97e362c032911dbc8f:2)

```
so I assume it trusts this response blindly, and sets `serverUrl` to  `ManualAddress` regardless of whether it's actually reachable?

I tried adding a trailing slash to the 'Base URL' entry under 'Networking' but that is always stripped on save, so maybe this is the underlying issue? I checked and I should be on the latest image `docker.io/jellyfin/jellyfin@sha256:1694ff069f0c9dafb283c36765175606866769f5d72f2ed56b6a0f1be922fc37`

As a temporary workaround of sorts (if you don't want to change the routing), making `~/.config/jellyfin-desktop/settings.json` read-only seems to work okay:
```sh
chmod a-w ~/.config/jellyfin-desktop/settings.json
```

PR #225, opened 2026-04-20
Fetch dependencies using cmake

BrainTeazer:
Fetching dependencies via CMake is, in my opinion, a better approach than submodules and copying the source files from the dependencies. This would allow adding other dependencies much more easily, and would reduce some of the source code clutter. Some things to consider:

- [ ] CEF builds with MSVC, but libmpv uses MSYS/MINGW to build. So, right now the windows build is a bit of a mess. I think if someone has an idea for adjusting the build such that they all use either MSVC or MSYS, then it would reduce the complexity of the windows build.
- [ ] I think CEF could also be pinned to a specific version and updated semi-regularly using the renovate bot. This would also allow fetching CEF via CMake. Otherwise, executing the download_cef.py file from CMAKE could be explored.
- [ ] The build servers pass, but I do not have a windows machine to test manual building.

Tested for:
- [x] Linux (Wayland)
- [x] MacOS (ARM)
- [ ] Windows


andrewrabert:
No thanks,I prefer vendored.

Issue #201, opened 2026-04-18
audio in live tv is broken


bendschs:
**Describe the bug**

audio for live tv streams is broken, no sound at all. the stream seems to be available (see screenshot of playback info), but sound is not played.

**Steps to reproduce**
1. play live tv channel
2. no sound

**System information**
 - Installation Method: mac os silicone

**Additional context**
Screenshots, logs, or any other relevant information.
<img width="239" height="266" alt="Image" src="https://github.com/user-attachments/assets/31ae261e-89be-417e-92da-b82f9c760aac" />

PR #233, opened 2026-04-21
Default to auto instead of disabled for audio stream.

iamfil:
This change fixes audio on LiveTV. When using LiveTV, DefaultAudioStreamIndex is -1 which results in no audio.


iamfil:
This fixes https://github.com/jellyfin/jellyfin-desktop/issues/201 .

andrewrabert:
Thanks!

PR #236, opened 2026-04-23
Add AV1 transcoding support and fix HLS audio track selection

### LINKED ISSUE #89: AV1 codec transcoding not working
**Describe the bug**
Transcoding to AV1 never happens, only x264/HEVC (on both linux and macbook).

**Steps to reproduce*
1. Enable AV1 encoding in the jellyfin server transcode options
2. Force transcoding of any source (x264, x265, AV1) via bandwidth setting

**System information**
 - Installation Method: truenas app
 - OS: Fedora 43
 - GPU: AMD RX 7900XTX
 - (Linux Only) **X11** or **Wayland**? wayland

**Additional context**
Screenshots, logs, or any other relevant information.

cebarks:
## Summary
This PR enables AV1 transcoding support. The changes required to implement this are as follows:
- Swap HLS transcoding container from MPEG-TS to fMP4 (`mp4`). This is required as MPEG-TS containers do not support AV1 video streams.
- When using fMP4 for the HLS container, the server only includes the selected audio track (always `aid=1`) in the stream, so requesting `aid=2+` resulted in silent audio when transcoding. Switching to auto track selection for HLS URLs solves this since the server already selects the correct audio track to include in the stream.

Fixes #89

## Test plan
- [x] Play content that forces transcoding (lower bitrate limit) — verify AV1 appears as the video codec in playback info
- [x] Select a non-primary audio track (e.g. 5.1 AC3 on stream index 2) during transcoding — verify audio plays correctly
- [x] Verify surround sound (5.1/7.1) works during both direct play and transcoding
- [x] Verify stereo content still works during transcoding
- [x] Test direct play (no transcoding) is unaffected

🤖 Generated with help from [Claude Code](https://claude.com/claude-code) (Claude Opus 4.6). Reviewed by a real human.

### DIFF
--- a/src/browser/web_browser.cpp
+++ b/src/browser/web_browser.cpp
@@ -143,8 +143,9 @@ bool WebBrowser::handleMessage(const std::string& name,
                  audioIdx, subIdx, startMs, url.c_str());
         MpvHandle::LoadOptions opts;
         opts.startSecs = startMs / 1000.0;
-        opts.audioTrack = audioIdx;
-        opts.subTrack = subIdx;
+        bool isHls = url.find("master.m3u8") != std::string::npos;
+        opts.audioTrack = isHls ? MpvHandle::kTrackAuto : audioIdx;
+        opts.subTrack = isHls ? MpvHandle::kTrackAuto : subIdx;
         g_mpv.LoadFile(url, opts);
     } else if (name == "playerStop") {
         g_mpv.Stop();
--- a/src/web/native-shim.js
+++ b/src/web/native-shim.js
@@ -370,7 +370,7 @@
                     Container: 'mp4',
                     Type: 'Video',
                     Protocol: 'hls',
-                    AudioCodec: 'aac,ac3,eac3,opus,flac,vorbis',
+                    AudioCodec: 'aac,mp3,ac3,eac3,opus,flac,vorbis',
                     VideoCodec: 'av1,h264,h265,hevc,mpeg4,mpeg2video',
                     MaxAudioChannels: '6'
                 },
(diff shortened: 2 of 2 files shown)


gnattu:
> Swap HLS transcoding container from MPEG-TS to fMP4 (mp4). This is required as MPEG-TS containers do not support AV1 video streams.

mpv does not work well with current jellyfin's fmp4 transcoding last time I checked and this will cause regressions on stream seeking and/or actual streaming

cebarks:
hmm, from my testing I wasn't having any issues when trascoding, but im not sure i tested seeking during a stream. Normal streaming without transcoding did work fine however. I tested on macOS and linux, but didn't have a windows machine. the only problem i ran into was audio  stream selection issues but that was due to a bug in the stream selection logic in the client, not due to transcoding specifically.

gnattu:
>  but im not sure i tested seeking during a stream

That is the part caused most issues. Especially when transcoding only happens to video and audio is coping.

andrewrabert:
I'll be handling both track selection and supported codecs in https://github.com/jellyfin/jellyfin-desktop/pull/285

cebarks:
will #285 enable av1 as a transcoding target?

gnattu:
Enable av1 as transcoding target is the trivial part after ensuring fmp4 compatibility. If the seeking issues persists we have to stick with mpegts until that is resolved first.

I just checked and can confirm the compatibility issue still persists. Seeking to a future position that is not yet ready for transcoding would outright crash the playback with fmp4 and the same does not happen with mpegts

nyanmisaka:
For the record, there's a WIP av1 over mpegts impl in ffmpeg https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/21307

Official specs from AOM: https://github.com/AOMediaCodec/av1-mpeg2-ts

PR #239, opened 2026-04-23
wayland: guard wp_viewport_set_source against committed buffer size (fixes crash on KWin)

bogdan-velicu:
## Problem

On KWin/Wayland, the app crashes with `VK_ERROR_SURFACE_LOST_KHR` whenever the window undergoes a size transition where the OSD dimensions oscillate (e.g. snapping to half-width via a tiling shortcut and back, or the normal startup sequence where OSD briefly reports a smaller size before settling).

The crash is reproducible on Arch Linux / KDE Plasma 6 by:
- Pressing Win+F repeatedly to toggle window snap
- Simply launching the app (the startup OSD sequence 950 → 1908 is enough to trigger it)

### Crash log
```
[CEF] wp_viewport#70: error 2: source rectangle x=0,y=0,w=1908,h=1018 extends outside of the content area x=0,y=0,w=950,h=1018
[mpv] vo/gpu-next/libplacebo: Failed presenting to queue …: VK_ERROR_SURFACE_LOST_KHR
[mpv] vo/gpu-next: Failed presenting frame!
[mpv] vo/gpu-next/wayland: Error occurred on the display fd
```

## Root cause

**KWin validates `wp_viewport_set_source` against the last non-null committed buffer at message-processing time** — not at `wl_surface.commit` time as the Wayland viewporter spec implies. Setting a source rectangle larger than the committed buffer raises `wp_viewport error 2 (out_of_buffer)`, which terminates the Wayland display connection and causes `VK_ERROR_SURFACE_LOST_KHR` in mpv's Vulkan path.

Confirmed with `WAYLAND_DEBUG=1`:

```
-> wp_viewport_set_source(0, 0, 1908, 1018)   ← committed buffer is 950-wide
{Display Queue} wl_display#1.error(wp_viewport#70, 2,
    "source rectangle x=0,y=0,w=1908,h=1018 extends outside of the
     content area x=0,y=0,w=950,h=1018")
```

### Why the mismatch occurs

There are multiple call sites that set the viewport source to the *current mpv window size* (`mpv_pw`) without checking whether `wl_present` has yet delivered a buffer of that size:

1. **`update_surface_size_locked` shrink path** — sets `source = pw` assuming the committed buffer is ≥ `pw`, but if `mpv_pw` previously grew (via a grow OSD event) without a matching `wl_present`, the committed buffer can be smaller than `pw`.
2. **`wl_present` / `wl_overlay_present` / etc.** — sets `source = min(w, mpv_pw)`, which can exceed the committed buffer size when the two are out of sync.
3. **`wl_end_transition_locked`** — set `source = mpv_pw` unconditionally, regardless of committed buffer size.

KWin also **retains the last non-null buffer as its validation reference** even after a null buffer is committed (e.g. in `wl_begin_transition_locked`), so resetting our tracking to 0 on null commits was wrong.

## Fix

Add `set_viewport_source_clamped()`, a helper that:
- sets the explicit source when `(cw, ch)` fits within the last non-null committed buffer size, or when no buffer has ever been committed;
- otherwise emits the **disabled form** (`-1, -1, -1, -1`), which is always valid and tells the compositor to use the full buffer extent (visually correct for one transitional frame until `wl_present` delivers the correctly-sized buffer).

Track the last non-null committed buffer size per subsurface (`cef_committed_w/h`, `overlay_committed_w/h`, `about_committed_w/h`). Null commits do **not** reset these fields for the reason above.

Apply the guard at every `wp_viewport_set_source` call site on subsurfaces whose source size may exceed the committed buffer. The popup subsurface is intentionally not tracked — its source is always set together with a freshly created buffer of matching size.

## Testing

- **Environment:** Arch Linux, KDE Plasma 6.x, KWin/Wayland, AMD GPU (RADV)
- **Verified with `WAYLAND_DEBUG=1`:** no `wp_viewport` errors after the fix
- **Stress-tested:** repeated Win+F window snap toggling, normal app startup — no crashes
- **Unaffected paths:** fullscreen toggle, overlay show/hide, about panel — all working normally

### DIFF
--- a/src/platform/wayland.cpp
+++ b/src/platform/wayland.cpp
@@ -94,6 +94,19 @@ struct WlState {
     bool transitioning = false;
     bool was_fullscreen = false;
 
+    // Size of the last non-null committed buffer for each clamped subsurface.
+    // KWin validates wp_viewport_set_source against the last non-null committed
+    // buffer at message-processing time (not at wl_surface.commit time), and
+    // retains that size even after a null buffer is committed.  We track these
+    // so set_viewport_source_clamped can fall back to source=-1 (full-buffer,
+    // always valid) instead of risking a protocol error when the proposed
+    // source exceeds the committed buffer.  0 means no buffer has ever been
+    // committed on that surface.  The popup subsurface is not tracked because
+    // its source is only set together with a new buffer of matching size.
+    int cef_committed_w = 0,     cef_committed_h = 0;
+    int overlay_committed_w = 0, overlay_committed_h = 0;
+    int about_committed_w = 0,   about_committed_h = 0;
+
 #ifdef HAVE_KDE_DECORATION_PALETTE
     org_kde_kwin_server_decoration_palette_manager* palette_manager = nullptr;
     org_kde_kwin_server_decoration_palette* palette = nullptr;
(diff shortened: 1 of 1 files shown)


andrewrabert:
I can't say I've seen this issue on Plasma 6 Wayland (amdgpu).

andrewrabert:
I'm going to close this since for now since I'm not sure it's needed. It also resulted in the CEF content sometimes stretching during resize - something I consider a hard [RFC 2119 MUST NOT](https://datatracker.ietf.org/doc/html/rfc2119#section-2).

If you're having issues, please open an issue and I'd be happy to look more into the problem you're having. (also - including a video capture of what you're seen would be great!)

RayJW:
> I can't say I've seen this issue on Plasma 6 Wayland (amdgpu).

This build fixes an issue for me, where going full-screen in Sway with the new desktop application simply makes the window disappear without a way to bring it back. I just tested the latest build from main and that issue is still happening there.

PR #244, opened 2026-04-24
fix external subs not loading initially

trbn1:
this fixes external subs not loading initially

### DIFF
--- a/src/web/mpv-video-player.js
+++ b/src/web/mpv-video-player.js
@@ -158,7 +158,7 @@
                 }
 
                 // Convert subtitle index to relative
-                let subParam = MpvPlayerCore.TRACK_DISABLE;
+                let subParam = MpvPlayerCore.TRACK_AUTO;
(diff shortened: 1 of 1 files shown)


andrewrabert:
I need to test this due to use of `auto`. I have concerns with multi-track containers where a track is set to forced or default.

trbn1:
This shouldn't change the existing functionality. The issue here is that currently by setting TRACK_DISABLE we set `mpv sid=0` for the case where we start with external subs, which takes priority later when we try to `mpv sub-add <url> select` the subtitles, but subs are force disabled by `sid=0` at that point. And this is why right now swapping between external (default `sid=0`) -> internal (sets `sid=auto`) -> external works as a workaround.

andrewrabert:
mpv auto track selection is indeed very problematic as I just encountered a separate issue. This will be handled by https://github.com/jellyfin/jellyfin-desktop/pull/285

PR #257, opened 2026-04-27
feat(wayland): add client-side window controls

### LINKED ISSUE #141: Add client-side decoration (CSD) support
**Describe the feature**
Add client-side decoration (CSD) support.
- Needed by GNOME to have window controls
- Can allow the full-window to display the video.
- Should be automatically enabled for window managers lacking support for basic window controls (ie. enabled when GNOME, disabled when KWin)
- Should have an option to enable regardless (so that KWin doesn't attach a titlebar + buttons; also macOS and Windows)
- Must not overlap web content.
- Must disappear/reappear when the OSD does so (configurable).

Relevant: https://github.com/mpv-player/mpv/issues/11665

AlfEspadero:
## Summary
 
 Implements client-side window controls for Wayland sessions so users still have **minimize**, **maximize**, and **close** even when compositor-provided decorations are missing (e.g. 
GNOME).
 
 ### What changed
 
 - Added Wayland client-side control buttons in `native-shim.js`:
   - Minimize
   - Toggle maximize
   - Close
 - Wired new native IPC handlers in `WebBrowser`:
   - `windowMinimize`
   - `windowToggleMaximize`
 - Added mpv wrappers for window state updates:
   - `window-minimized`
   - `window-maximized`
 - Passed active display backend into renderer script injection:
   - Set `JFD_DISPLAY_BACKEND` in `main.cpp`
   - Injected `__DISPLAY_BACKEND__` in `cef_app.cpp`
 - Made controls Wayland-only and hidden while fullscreen.
 
 ## Why
 
 On some Wayland environments (notably GNOME), server-side decorations are not reliably available for this app, which leaves users without window controls. This adds a client-side 
fallback so window management remains accessible.

Full disclosure, I did this with the help of AI solely because I just wanted the app to work locally so I could use it in my day-to-day. But if it helps the overall project I'd be chuffed.
 
 ## Scope / behavior
 
 - **Wayland:** shows client-side controls in-app.
 - **X11/macOS/Windows:** unchanged.
 
 Closes #141

### DIFF
--- a/src/browser/web_browser.cpp
+++ b/src/browser/web_browser.cpp
@@ -90,6 +90,7 @@ CefRefPtr<CefDictionaryValue> WebBrowser::injectionProfile() {
         "notifyRateChange",
         "appExit", "setSettingValue", "themeColor",
         "setOsdVisible", "setCursorVisible", "toggleFullscreen",
+        "windowMinimize", "windowToggleMaximize",
         "menuItemSelected", "menuDismissed",
     };
     static const char* const kScripts[] = {
@@ -187,6 +188,10 @@ bool WebBrowser::handleMessage(const std::string& name,
         }
     } else if (name == "toggleFullscreen") {
         g_platform.toggle_fullscreen();
+    } else if (name == "windowMinimize") {
+        g_mpv.SetWindowMinimized(true);
+    } else if (name == "windowToggleMaximize") {
+        g_mpv.SetWindowMaximized(!mpv::window_maximized());
     } else if (name == "saveServerUrl") {
         std::string url = args->GetString(0).ToString();
         Settings::instance().setServerUrl(url);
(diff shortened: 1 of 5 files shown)


andrewrabert:
Thanks for taking a stab at this, but I'm going to have to close this due to being low-quality.

1. It results in CSD on KWin/Wayland where there are already decorations.
2. The web assets for the buttons shouldn't know/care about whether Wayland is used; that's a poor separation of concerns that unnecessarily couples CSD to a specific display protocol. I've clarified the issue that this _should_ be something usable on all platform. 
3. The buttons overlap the web content. There's a macOS-specific injection elsewhere in the codebase that may be useful here.
4. The buttons remain visible and overlap video content even when the OSD disappears.
5. The buttons are implemented on the  wrong layer and don't appear on the overlay layer (not saying that's the correct layer either. Probably should be a new, separate CEF layer)


Screenshot_20260427_193119

image

image



---

If you're going to open another PR, I require you post screenshots next time.

AlfEspadero:
fair enough, thank you for the feedback

PR #264, opened 2026-04-29
ci: guard macOS Monterey deployment target (#248)

### LINKED ISSUE #248: macOS builds require macOS 15.0 — should support macOS 12 (Monterey)
The macOS builds (both nightly and release) set an unnecessarily high deployment target, making them unusable on macOS 12 (Monterey) and macOS 13 (Ventura).

The nightly x86_64 build from https://nightly.link/jellyfin/jellyfin-desktop/workflows/build-macos/main/macos-x86_64.zip requires macOS 15.0, which excludes some Intel Macs (Monterey is the last supported macOS for most Intel hardware).

## Root cause

`MACOSX_DEPLOYMENT_TARGET` is not explicitly set in the CI workflow. The build inherits the deployment target from the GitHub Actions runner OS, which is currently macOS 14 or 15. This propagates into all compiled binaries via the `minos` field in `LC_BUILD_VERSION`.

The same issue affected the archived jellyfin-media-player (Qt-based client).

## Evidence

I successfully built jellyfin-media-player v2.0.0 (Qt6) with `CMAKE_OSX_DEPLOYMENT_TARGET=12.0` and confirmed it runs on macOS 12.7.6 (Intel iMac). The application code itself does not use any macOS 13+ APIs.

The only source-level fix needed was in mpv's `osdep/mac/swift_compat.swift` (a one-line version check adjustment). Everything else was purely a build configuration issue.

Working build available at: https://github.com/ctismer/jellyfin-desktop-qt/releases/tag/v2.0.0-macos12

## Suggested fix

Add `MACOSX_DEPLOYMENT_TARGET: "12.0"` as an environment variable in the macOS build workflow. This should be sufficient unless the CEF binaries themselves require a higher target.

## Affected users

All Intel Mac users stuck on macOS 12 Monterey (iMac 2015-2017, MacBook Pro 2015-2016, Mac Pro 2013, Mac mini 2014, etc.)

alceops:
## Summary
- set the macOS CI/build deployment target to Monterey (`12.0`) for the app and CMake configure step
- pass the same deployment target into the vendored mpv Meson setup/compile environment
- add a pre-DMG bundle guard that inspects Mach-O `minos` values with `vtool` so CI catches dylibs/frameworks that drift above the target

## Verification
- `git diff --check`
- `python3 -m py_compile dev/tools/verify_macos_deployment.py`
- static workflow/CMake assertions for deployment target propagation and the verification step

I could not run a full macOS/Xcode/Homebrew/CEF/DMG build in this Linux cron environment, so this PR is a CI/build-system guard first slice for #248 rather than a claim that the Monterey runtime crash is fully fixed.

Fixes #248

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -24,6 +24,9 @@ jobs:
           - runner: macos-15-intel
             arch: x86_64
     runs-on: ${{ matrix.runner }}
+    env:
+      MACOSX_DEPLOYMENT_TARGET: '12.0'
+      CMAKE_OSX_DEPLOYMENT_TARGET: '12.0'
 
     steps:
       - name: Checkout
@@ -65,12 +68,19 @@ jobs:
       - name: Configure and build
         run: |
           cmake -B build -G Ninja \
-            -DCMAKE_BUILD_TYPE=Release
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_OSX_DEPLOYMENT_TARGET="$MACOSX_DEPLOYMENT_TARGET"
           cmake --build build
 
       - name: Create app bundle
         run: |
           cmake --install build --prefix build/output
+
+      - name: Verify Monterey deployment target
+        run: |
+          python3 dev/tools/verify_macos_deployment.py \
+            "build/output/Jellyfin Desktop.app" \
+            --target "$MACOSX_DEPLOYMENT_TARGET"
 
       - name: Create DMG
         run: |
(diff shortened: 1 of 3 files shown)


alceops:
CI did exactly what the guard was meant to do: both macOS legs now fail at the new verifier because bundled Homebrew-built dependencies still advertise newer `LC_BUILD_VERSION` minimums, even with `MACOSX_DEPLOYMENT_TARGET=12.0` set for the app/mpv build.

Examples from run 25094179850:
- arm64: bundled dylibs report `minos 15`
- x86_64: most bundled dylibs report `minos 14`, with `libcrypto`, `libssl`, `libpcre2`, and `libvulkan` at `minos 15`

That means this PR is currently a diagnostic guardrail, not a complete Monterey fix. I am not going to loosen the check to make it green, because that would recreate the false-positive build. The next useful slice would be replacing the macOS Homebrew bottle path for the offending bundled dependencies with source builds or a Monterey-compatible dependency bundle, then keeping this verifier as the acceptance gate.

andrewrabert:
Erm - this isn't actionable. Please reopen once you have something working.

PR #269, opened 2026-04-30
Add macOS 12 Monterey support

### LINKED ISSUE #248: macOS builds require macOS 15.0 — should support macOS 12 (Monterey)

ctismer:
Restores building for macOS 12 Monterey (last macOS for Intel Macs).
Closes #248. Tested on macOS 12.7.6 (Intel).

## Changes

- CI: matrix builds for arm64 (`macos-15`) and x86_64 (`macos-15-intel`)
  with `MACOSX_DEPLOYMENT_TARGET=12.0`. ffmpeg 8.1 and libplacebo 7.360.1
  built from source against the 12.0 SDK (Homebrew bottles target 13+).
- mpv: enable the `NSLocking.withLock` backport for Swift 5.7 (the
  version guard was off-by-one) and pin mpv's Swift compile to
  `<arch>-apple-macos12.0` so libmpv doesn't reference Swift runtime
  symbols missing on Monterey (`__StaticArrayStorage`).
- `cmake/FindCEF.cmake`: forward sysroot/deployment-target/compilers to
  the CEF sub-build.
- `cmake/CompleteBundleMac.cmake.in`: search both `/usr/local` and
  `/opt/homebrew` for MoltenVK; generalize bundling for non-system
  absolute paths.
- `src/platform/macos.mm`: drop explicit `[NSApp finishLaunching]` —
  `[NSApp run]` does it; calling twice crashes `_createMenuRef` on 12.

Spotify CEF binaries themselves already declare `minos 12.0`.

Build time grew to ~18 min because of the from-source ffmpeg/libplacebo;
caching those into a separate prebuilt artifact would be a natural
follow-up if it matters.


andrewrabert:
The build time increase of ~18 min is pretty rough, but I guess that's unavoidable. 

I'm thinking this should be a secondary build type (legacy?) so the existing macos workflows and artifacts remain as-is. Both the build-time increase and age of the OS this is targetting has me concerned for long-term viability. Also not sure the arm variant needs support as all arm macs still support the latest OS.

What do you think?

andrewrabert:
Side note - I deeply appreciate this. I don't use macOS as my main OS so I appreciate community contributions (especially for lesser used hardware like the older macs).

ctismer:
Sounds good — I'll split this into a separate `build-macos-legacy.yml` that only does x86_64-Monterey, leave the existing `build-macos.yml` untouched, and drop the arm64 variant. The non-workflow fixes (NSCarbonMenuImpl, MoltenVK search, FindCEF Xcode forwarding) are harmless on newer macOS so I'll keep those in the shared paths.

For the 18 min build time: I'd be happy to follow up with a separate PR that caches the from-source ffmpeg + libplacebo via actions/cache, which should bring warm runs down to ~5–7 min. Keeping that out of this PR to keep the diff focused.

Also: thanks for the kind words. Happy to keep this branch alive for the holdouts (my main is M1, but the Jellyfin-Mac in the living room is an older MBP stuck on Monterey).

ctismer:
Refactored as discussed: there's a new build-macos-legacy.yml that targets x86_64-Monterey only, while build-macos.yml is now byte-identical to upstream. The source-tree changes are reduced to what's actually needed — the arch-aware MoltenVK search and removal of the redundant -[NSApp finishLaunching] call. Tested on macOS 12.7.6 (Intel iMac) and macOS 14 (M1).

ctismer:
Folded the caching in here too — turned out to be ~10 lines of YAML plus moving the install prefix off /usr/local, so a separate PR didn't seem worth the round-trip. Warm CI runs are now ~7 min instead of ~18.

Issue #76, opened 2026-03-30
Add option to disable KWin titlebar themeing


enter-a-random-username:
**Describe the bug**
App has black bar and not same color as rest of system

**Steps to reproduce**
1. Start jellyfin-deskop

**System information**
 - Installation Method: built from source
 - OS: Linux KDE Wayland

andrewrabert:
Need screenshot.

enter-a-random-username:
<img width="322" height="31" alt="Image" src="https://github.com/user-attachments/assets/1ebd3026-70ba-4af6-be23-b5be64a37d06" /> focused


<img width="322" height="31" alt="Image" src="https://github.com/user-attachments/assets/76f9d73d-e530-4494-b4e3-bce6503d65c3" /> off focus

ALL other software

<img width="337" height="34" alt="Image" src="https://github.com/user-attachments/assets/18dfadbe-f2db-4ab1-a517-52e8db5be8b1" /> focused

<img width="230" height="36" alt="Image" src="https://github.com/user-attachments/assets/a5ed24b6-d2b2-458b-aee9-ff152a504af1" /> off focus

Chromium uses the same decoration as kate, always the same colors except this one application

andrewrabert:
Give me a full window screenshot please.

enter-a-random-username:
I want the kwin title bar styling, but the colors are off, just dark gray and done  (only thing that changes is font color), not the system style that is configured.

<img width="1826" height="1180" alt="Image" src="https://github.com/user-attachments/assets/abff9c11-b871-4855-8d78-fbe65a3c6858" />

---

Every other application on focus
<img width="1147" height="96" alt="Image" src="https://github.com/user-attachments/assets/4f8e85b7-1805-4383-a534-deed16e23692" />

andrewrabert:
An option is now available to disable the default titlebar theming https://github.com/jellyfin/jellyfin-desktop/pull/280

PR #280, opened 2026-05-01
Toggle titlebar theming

rodis120:
Added option to disable titlebar color theming on wayland. When disabled it uses current system theme.

Related issue: #76 

Settings:
image

Examples:
image
image
image


andrewrabert:
Thanks!

PR #284, opened 2026-05-02
fix macOS HDR handling

### LINKED ISSUE #241: HDR is not as bright as it is in other players
### Confirmation

### Version

3.0.0-dev+03ead07

### Platform

macOS

### OS version

26.3.1

### GPU

Macbook pro 14 M1

### Describe the bug

HDR in safari, IINA seems much brighter than Jellyfin player.

### Steps to reproduce

1. Play a HDR film in Safari
2. Play same HDR film in Jellyfin
3. Check the difference

### Additional context

_No response_

gnattu:
on macOS, we need trigger HDR mode of the display by:
- target-colorspace-hint=yes
- pulse mpv geometry to activate HDR swapchain

I hope there is better way to do this but it seems like the only reliable way to force mpv recreate a swapchain is geometry resize. Without a forced swapchain recreation, the swapchain never request EDR headroom and will play HDR video effectively in tonemapped SDR.

target-colorspace-hint=auto will not work with Apple branded HDR displays as Apple displays are dual-mode that can turn on HDR automatically when the application want to display HDR content. target-colorspace-hint=auto would only detect the SDR mode of the display and never request HDR.

Fixes #241

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -126,6 +126,63 @@ static void log_mpv_message(const mpv_event_log_message* msg) {
     }
 }
 
+#ifdef __APPLE__
+static std::atomic<bool> s_edr_geometry_pulse_pending{false};
+static std::atomic<int> s_edr_geometry_original_w{0};
+static std::atomic<int> s_edr_geometry_original_h{0};
+static std::atomic<int> s_edr_geometry_pulse_w{0};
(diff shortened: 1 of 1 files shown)


gnattu:
Closing as this would not work for full screen, as mpv will ignore geometry changes in full screen

Issue #210, opened 2026-04-19
No Sound when Playing Videos


Terrorwolf01:
**Describe the bug**
As above, when I try to play Videos through the newest Jellyfin-Desktop, I have no Audio. Video is playing without Problem tho. This doesn't affect a Music Library tho. This happens regardless of Transcoding, while working in Firefox without Problems.

**Steps to reproduce**
1. Play something

**System information**
 - Installation Method: Installed via AUR -git Package, just updated to latest commit as of writing. (I installed it about 2 weeks and have this bug since then already tho)
 - OS: EndeavourOS
 - GPU: NVIDIA RTX 4090
 - (Linux Only) **X11** or **Wayland**? Wayland

**Additional context**
Screenshots, logs, or any other relevant information.
Installed Plugins (Without Metadata Plugins):  Chapter Segments Provider, Collection Sections, Custom Tabs, File Transformation, Home Screen Sections, Jellyfin Enhanced, Jellyfin Tweaks, LDAP-Auth, Media Bar, Merge Versions, Playback Reporting, Plugin Pages, Reports, Session Cleaner, SkinManager, SSO-Auth, Subtitle Extract, Webhook

Happy to provide more Infos or how some settings are set.

daniyl-x:
Same issue with the old client on X11, installed from flathub.

andrewrabert:
This should be completely fixed as of #285 and https://github.com/jellyfin/jellyfin-desktop/commit/8b026165c0991a2da678c12b12941f4e73e7dc85

Issue #175, opened 2026-04-15
wrong audio or sub


TIGamingTV:
**Describe the bug**
Sometimes jellyfin uses the wrong audio stream. 
I selected english as the preferred language but sometimes in rare cases it selects italian or something else. 

With subtitles its more often. 
I selected english and only forced. Sometimes it loads italian forced subs or english (not forced) subs. 

I tried to find similarities but there were none. It is random and rare.

**System information**
 - Installation Method: 13 April
 - OS: Windows 11
 - GPU: RX 9070 XT

**Additional context**
Screenshots, logs, or any other relevant information.

enter-a-random-username:
I noticed that as well. Switched subs, but the MKV flags stayed the same as in files before.
(Similar rare thing is it starts the same episode if I skip a preview chapter or so.)

Never had that issue in Web/Jellyfin-desktop (old) or 3rd-Party

andrewrabert:
Completed in https://github.com/jellyfin/jellyfin-desktop/pull/285

Issue #147, opened 2026-04-12
Playback next-episode if set to non-default, doesn't correctly update before starting to play content


NekoShinobi:
**Describe the bug**

Essentially when jellyfin starts the next episode, it'll reuse the settings from the last episode. Jellyfin-desktop just plays the default no matter what, so despite audio/sub being marked as something else, the video just plays the default audio/sub

For example,
I'm playing a video with Japanese / English audio. English is the default audio, but I select Japanese for the episode. When playing the next video, jellyfin will auto select Japanese from the last video, but the video will play English still.

**Steps to reproduce**
1. Play dual audio/multi-sub file
2. select something other than default
3. go next video
4. jellyfin auto selects the same audio/sub you selected last video, but the video still plays whatever is the default.

**System information**
 - Installation Method: AUR jellyfin-desktop-git
 - OS: EndeavourOS (Linux 6.19.10)
 - GPU: NVIDIA RTX 4090
 - Wayland (Hyprland 0.54.3)

**Additional context**
Screenshots, logs, or any other relevant information.

NekoShinobi:
This seems to be kind of a flaky behavior. I was able to reproduce a few times, but I wasn't able to consistently reproduce it.

Mathis0d:
The bug consistently occurs it just appears when you change the audio track or subtitles while you’re already watching an episode, not when you select them before starting playback.

Additionally, if you switch episodes a second time, it correctly remembers the change you made the bug only happens on the first episode change.

Issue #116, opened 2026-04-06
Add force transcoding options under client settings


JIKSse:
**Describe the feature**
Adding more granular client settings options that were previously in old clients. I found the force transcode: 4k, av1, hevc, hdr etc options especially useful on older clients that dont have hardware decoding support for newer codecs and/or have weaker CPUs/GPUs that cant handle 4K/HDR playback well enough. 

**Did Jellyfin Media Player / the Qt desktop client have this feature?**
Yes 

**Additional context**

<img width="798" height="681" alt="Image" src="https://github.com/user-attachments/assets/1467291a-8ac8-41ab-aafc-7a227b880538" />

andrewrabert:
Completed in https://github.com/jellyfin/jellyfin-desktop/pull/285

PR #285, opened 2026-05-02
mpv track-related fixes

andrewrabert:
Disable all automatic track selection in mpv. It's problematic and can conflict with the expectations of jellyfin-web. It has historically been the cause of various bugs dating back to the Qt 5 desktop app.

Fixes:
- https://github.com/jellyfin/jellyfin-desktop/issues/89
- https://github.com/jellyfin/jellyfin-desktop/issues/116
- https://github.com/jellyfin/jellyfin-desktop/issues/147
- https://github.com/jellyfin/jellyfin-desktop/issues/175
- https://github.com/jellyfin/jellyfin-desktop/issues/210


src/jellyfin/device_profile.cpp
```diff
@@ -0,0 +1,262 @@
...
+    // Hardcoded transcoding fallback. These describe what the server should
+    // produce when direct play isn't possible; the codec lists are the
+    // server's preference order.
+    static constexpr const char* kTranscodeContainer        = "ts";
+    static constexpr const char* kTranscodeProtocol         = "hls";
+    static constexpr const char* kTranscodeMaxAudioChannels = "6";
+    static const std::vector<std::string> kTranscodeVideoCodec =
+        {"av1", "hevc", "h264", "mpeg4", "mpeg2video"};
+    static const std::vector<std::string> kTranscodeAudioCodec =
+        {"opus", "aac", "mp3", "ac3", "vorbis", "flac"};
```
nyanmisaka: TS has limited codec support.
For video, the server can only handle h264 and hevc. For audio, there are limitations: https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Model/Dlna/StreamBuilder.cs#L31-L33

andrewrabert: Didn't realize that. Nice find.

PR #287, opened 2026-05-03
macos: don't scale precise scrolling deltas

seridescent:
fixes the unexpectedly sensitive scrolling when using precise scrolling devices like the magic trackpad. preserves existing behavior for imprecise devices.

i gave this a sanity check using both a trackpad and a mouse that i confirmed provide precise and imprecise scrolling deltas, respectively.

i like the app so far! the scrolling was just bugging me.

### DIFF
--- a/src/input/input_macos.mm
+++ b/src/input/input_macos.mm
@@ -383,11 +383,15 @@ static void fill_key_event_from_nsevent(input::KeyEvent& e, NSEvent* event) {
 
 - (void)scrollWheel:(NSEvent*)event {
     NSPoint loc = [self mouseLocInView:event];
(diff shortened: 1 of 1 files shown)


gnattu:
Thank you for this but I have a more sophisticated solution to this: https://github.com/jellyfin/jellyfin-desktop/pull/294

The reason I did not push this is that I'm still traveling without a physical mouse that can test the non-precise scrolling path. If you could test that I would be very appreciated and I can mark that as non-draft

seridescent:
ah, okay. i'll comment on #294

andrewrabert:
Closed in favor of https://github.com/jellyfin/jellyfin-desktop/pull/294

PR #293, opened 2026-05-03
Set target-colorspace-hint=yes

### LINKED ISSUE #241: HDR is not as bright as it is in other players

gnattu:
This is necessary on macOS to properly trigger HDR mode as auto does cannot detect Apple's dual mode display which does not need explict HDR mode switch

This alone is not enough and we need https://github.com/andrewrabert/mpv/pull/2 to me merged first

Fixes #241

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -555,6 +555,9 @@ int main(int argc, char* argv[]) {
 
     g_mpv.SetHwdec(hwdec_str);
     g_mpv.SetOptionString("background-color", kBgColor.hex);
+#ifdef __APPLE__
+    g_mpv.SetOptionString("target-colorspace-hint", "yes");
+#endif
 
     // Restore saved window geometry. mpv's --geometry is always physical
     // pixels (m_geometry_apply at third_party/mpv/options/m_option.c:2296


andrewrabert:
Setting to draft until merge of https://github.com/andrewrabert/mpv/pull/2

andrewrabert:
This is still something I plan to revisit at some point. I'd like to save any mpv fork changes until after CEF rendering is fully stabilized.

andrewrabert:
The mpv fork will not be moving forward with regard to any feature changes. The mpv fork strictly a temporary aberration to get the bare minimum integration working with Jellyfin Desktop.

gnattu:
emmm so what are we going to do if it does not work with our use case and our use case is not something that upstream cares?

andrewrabert:
Then we'll reevaluate. I just want to see proof of trying to upstream before making any more fork changes.

TBH - I'm just burnout severely on this project and don't have the mental bandwidth to accommodate more fork changes. Especially with 3.0 stable still not a thing.

gnattu:
> TBH - I'm just burnout severely on this project and don't have the mental bandwidth to accommodate more fork changes. Especially with 3.0 stable still not a thing.

I get that, that was why I suggested to promote our mpv fork into jellyfin repo to make it something like jellyfin-mpv(just like jellyfin-ffmpeg), so that more maintainers can help, to reduce burnout.

> Then we'll reevaluate. I just want to see proof of trying to upstream before making any more fork changes.

I can try that but it would be hard because we are adding support for a non-exist upstream feature (gpu-next with libmpv. Maybe I can find a valid use case for this change to upstream but I need to investigate more.

andrewrabert:
> I get that, that was why I suggested to promote our mpv fork into jellyfin repo to make it something like jellyfin-mpv(just like jellyfin-ffmpeg), so that more maintainers can help, to reduce burnout.

I'd still like to see if the fork can be eliminated from Windows and/or macOS before 3.0 first. If the fork persists post 3.0, it'll be moved to jellyfin proper. 

> I can try that but it would be hard because we are adding support for a non-exist upstream feature (gpu-next with libmpv. Maybe I can find a valid use case for this change to upstream but I need to investigate more.

Keep in mind the usage on macOS and Windows is what I would equate to "get stuff done" vibing still. There _might_ be cleaner ways to achieve what's needed, but TBD. I've only recently gotten around to solidifying the Wayland approach in terms of mpv usage and still want to reevaluate what can be done with Win + mac.

PR #294, opened 2026-05-03
macos: improve scroll handling

### LINKED ISSUE #288: Scrolling is too fast / non-intertial
### Confirmation

### Version

3.0.0-dev+3d75025

### Platform

macOS

### OS version

15.7.1 (24G231)

### GPU

Apple M3 Max

### Describe the bug

When scrolling vertically or horizontally it scrolls very quickly, or at least that is what it feels like.  It also doesn't seem to have the usual macos inertial scrolling, so that may be what's causing the fast-scroll feeling, or it may be two separate issues.

### Steps to reproduce

Any scrolling on JD "Home" screen.

### Additional context

_No response_

gnattu:
This is the port of https://github.com/jellyfin/jellyfin-desktop/pull/99 to current native view, which handles both precise and non-precise scrolling in a more natural way.

Fixes https://github.com/jellyfin/jellyfin-desktop/issues/288

### DIFF
--- a/src/input/dispatch.cpp
+++ b/src/input/dispatch.cpp
@@ -133,7 +133,9 @@ void dispatch_scroll(const ScrollEvent& e) {
     auto b = active_browser();
     if (!b) return;
     CefMouseEvent me{};
-    me.x = e.x; me.y = e.y; me.modifiers = e.modifiers;
+    me.x = e.x; me.y = e.y;
+    me.modifiers = e.modifiers;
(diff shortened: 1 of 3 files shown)


seridescent:
i checked this out and confirmed that scrolling with an imprecise device (my mouse) works as expected given the patch. the scrolling has a sane sensitivity that seems to match other apps.

and +1 to the overly-sensitive precise scrolling being fixed.

PR #296, opened 2026-05-03
macos: render About in the active macOS browser

### LINKED ISSUE #291: macos native "About Jellyfin Desktop" not working properly
### Confirmation

### Version

3.0.0-dev+3d75025

### Platform

macOS

### OS version

15.7.1 (24G231)

### GPU

Apple M3 Max

### Describe the bug

Opening the macos-native "About Jellyfin Desktop" does not work properly.

I'd expect it to open a floating window with version information.  Instead it behaves possibly as if it has opened that window off-screen, but it does not appear even in the "show all windows" view (pressing F3) and I have to quit the app and restart to regain control.

Note this is *not* referring to the "About" window you get from right-clicking in the main UI, which does work as expected.

Image

### Steps to reproduce

1. Have JD in focus
2. Click "Jellyfin Desktop" next to the Apple menu at the top left of the screen
3. Click "About Jellyfin Desktop"

### Additional context

_No response_

gnattu:
This is a workaround for macOS's unreliable OSR browser rendering. The cross-browser rendering could fail after the about browser opened and display nothing.
The only workaround seems to be directly render with the active browser.

Fixes #291

I have to admit that this might over-complicated, but most behavior is bared on macOS and we already have similar workaround (single process CEF) for the platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -495,6 +495,7 @@ set(JS_SHIMS
     ${CMAKE_SOURCE_DIR}/src/web/mpv-audio-player.js
     ${CMAKE_SOURCE_DIR}/src/web/input-plugin.js
     ${CMAKE_SOURCE_DIR}/src/web/client-settings.js
+    ${CMAKE_SOURCE_DIR}/src/web/about.js
     ${CMAKE_SOURCE_DIR}/src/web/context-menu.js
 )
 set(EMBEDDED_JS_HEADER ${CMAKE_BINARY_DIR}/generated/embedded_js.h)
--- a/src/browser/about_browser.cpp
+++ b/src/browser/about_browser.cpp
@@ -1,7 +1,10 @@
 #include "about_browser.h"
 #include "app_menu.h"
 #include "browsers.h"
+#include "overlay_browser.h"
+#include "web_browser.h"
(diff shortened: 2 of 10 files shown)


andrewrabert:
Rendering the about dialog might be the solution to macOS's flaky About overlay. 

Not sure yet as I _might_ have found a way to force it to rerender via JS injection. The example is specific to improving rendering on resize, but similar may help with the About overlay. Seems hacky, but it's not an uncommon approach for CEF OSR https://github.com/chromiumembedded/cef/issues?q=is%3Aissue%20state%3Aopen%20requestAnimationFrame
```
window.addEventListener('resize', () => {
  requestAnimationFrame(() => requestAnimationFrame(() => {}));
});
```

I only have that locally for now and am not sure if I'll go with that yet for About yet. I'm currently refactoring the CEF layers with the intention of fixing a number of rendering issues (About included). At worst, I'll just have the About overlay render in the current active layer (Overlay or Web) like the context menus do. 

Will be handled in https://github.com/jellyfin/jellyfin-desktop/pull/349

---

side note - the recent refactors are not something I want to continue doing long-term, but am doing so as the application seems to be generally stabilizing across the platforms.

PR #299, opened 2026-05-03
remove broken fmp4 transcoding profile

gnattu:
mpv has compatibility issues with fmp4 streams; seeking in fmp4 HLS does not work. Seek to a not buffered section outright crashes playback — this does not happen with mpegts. Use mpegts for transcoding until fmp4 seeking is fixed in mpv.
This is not limited to Jellyfin fmp4 streams — even standard Apple test streams suffer the same seeking issue in mpv.

Regression from https://github.com/jellyfin/jellyfin-desktop/commit/e2f409a4f24eef8e135665431e82f854b7f7e1cd

### DIFF
--- a/src/jellyfin/device_profile.cpp
+++ b/src/jellyfin/device_profile.cpp
@@ -128,8 +128,6 @@ std::string Build(const mpv_capabilities::Capabilities& caps,
     // picks the most efficient target it can produce.
     static const std::vector<std::string> kTranscodeVideoCodec =
         {"av1", "hevc", "h264", "vp9"};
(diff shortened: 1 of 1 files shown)


andrewrabert:
Hm - works fine for me on Linux/Wayland. I assume you're seeing problems on macOS?

I really don't want to remove this as it would prevent me from transcoding+remuxing a bunch of audio codecs like opus and flac (https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Model/Dlna/StreamBuilder.cs#L31-L33)

I'm open to making profiles more configurable in settings, but let's discuss that in an issue.

PR #335, opened 2026-05-08
macOS legacy bundle: rewrite all non-system absolute dylib paths

ctismer:
Hotfix for a regression I introduced in eb025ed ("Cache ffmpeg and libplacebo builds"). That commit moved the build prefix from `/usr/local/jellyfin-deps` to `$HOME/jellyfin-deps` to avoid `actions/cache` permission problems when restoring under `/usr/local`. The bundler in `cmake/CompleteBundleMac.cmake.in` recognised dylib references via a small allowlist of prefixes (`@rpath`, `@loader_path`, `/opt/homebrew`, `/usr/local`) in two places: `fix_lib_dependencies` and `resolve_dependency`. Anything under `/Users/runner/...` matched neither, so libmpv's references to libavcodec/libplacebo/etc were left as absolute paths into the runner's home and the shipped DMG crashed at launch on the user's machine with `Library not loaded: '/Users/.../libplacebo.360.dylib'`.

The fix replaces both allowlists with a stable exclusion list — `/usr/lib`, `/System`, `/Library` are kept external; anything else absolute (or `@rpath`/`@loader_path`) gets bundled. Future build-prefix changes won't silently break bundling again.

Verified on macOS 12.7.6 Intel: DMG launches and runs; `otool -L` on the bundled libmpv shows ffmpeg/libplacebo refs now resolve to `@executable_path/../Frameworks/`.

### DIFF
--- a/cmake/CompleteBundleMac.cmake.in
+++ b/cmake/CompleteBundleMac.cmake.in
@@ -84,7 +84,11 @@ function(resolve_dependency dep_path out_var)
         list(GET cellar_candidates 0 result)
       endif()
     endif()
-  elseif(dep_path MATCHES "^(/opt/homebrew|/usr/local)")
+  elseif(dep_path MATCHES "^/")
+    # Any absolute path that exists on disk — covers Homebrew prefixes
+    # and custom build prefixes (e.g. $HOME/jellyfin-deps used by the
+    # legacy workflow). System paths are already filtered out by
+    # fix_lib_dependencies before we get here.
     if(EXISTS "${dep_path}")
       set(result "${dep_path}")
     endif()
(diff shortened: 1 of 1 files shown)


ctismer:
@andrewrabert hi again — sorry to show up so soon, but a bug crept in with the cache optimisation. The bundler's path-prefix allowlist in `CompleteBundleMac.cmake.in` didn't match our new `$HOME/jellyfin-deps` install prefix, so libmpv kept absolute references to the runner's home dir and the DMG crashed at launch on user machines. The change generalises both lookups to "any non-system absolute path" so future prefix changes can't silently break bundling again. Verified on macOS 12.7.6 Intel.

andrewrabert:
Nice catch, thanks!

ctismer:
Thanks for merging. For context — I actually set out to fix the display-sleep inhibit (#336) and only stumbled across this regression while trying to test that fix on Monterey. Shit happens.

Issue #245, opened 2026-04-25
Dolby Vision Profile 7 file outputs HDR10 instead of DV on LG G4 via HDMI


hausmanj:
### Confirmation

- [x] I have read the [Reporting Bugs](https://github.com/jellyfin/jellyfin-desktop/wiki/Reporting-Bugs) guide.
- [x] I acknowledge that the version of Jellyfin Desktop I'm reporting for starts with `3.0.0-dev`
- [x] I will disclose any use of custom styles or plugins that alter the web appearance.

### Version

3.0.0-dev+03ead07

### Platform

Windows

### OS version

Windows 11 25H2

### GPU

AMD Ryzen 7 PRO 6850H with Radeon Graphics, 3201 Mhz, 8 Core(s), 16 Logical Processor(s)

### Describe the bug

The 3.0 dev client is not passing Dolby Vision content, instead it is getting somehow converted to HDR.  Verified directly on my LG G4 and G5 tvs.

### Steps to reproduce

1.	Play Dolby Vison film or show with DV 7.6 in Jellyfin client 03ead07
2.	Verify video format output directly on tv
3.	Observe DV content somehow converted to HDR

### Additional context

Same DV content in client 1.12 plays Dolby Vision content fine.  The Properties window during playback in both clients indicated Dolby Vision and Direct Play, so it is not transcoding the video.  Multiple shows and movies tested with DV 7.6.

My tv flashes audio and video information at the top right of the screen.  Typically, I see Dolby Atmos for audio, then either HDR or Dolby Vision depending on what is being output.  On Jellyfin Desktop, I see Dolby Vision on the home page, but it changes to HDR whenever playing DV content.

I checked available settings in Jellyfin Desktop and did not find anything applicable.  Disabled hardware acceleration (long shot) but no change.  

Dolby Access and Microsoft Dolby Vision Extensions both installed

andrewrabert:
Probably an upstream mpv issue

hausmanj:
Update on this - the recent nightly build that enabled access to the MPV configuration folder allowed me to configure MPV to effectively pass Dolby Vision on windows 11.  I need to do this independently for each client, but its working on (2) windows clients so far.  The build is 267af8a from Action 424.  Here is the mpv.conf file that I'm using:

vo=gpu-next
gpu-api=d3d11
gpu-context=d3d11
hwdec=d3d11va-copy

target-colorspace-hint=no
target-trc=auto
target-prim=auto
target-peak=auto
[...]

PR #346, opened 2026-05-10
Add cross-platform mpv Lua script bundling

hausmanj:
Improves Dolby Vision handling during playback on Windows.  Resolves Issue #245

Changes:
- Adjusts Dolby Vision related playback handling
- Updates mpv/native playback interaction
- Keeps the change isolated to Windows playback behavior

Testing:
- Built successfully on Windows 11 25H2 x64, GPU AMD Ryzen 7 PRO 6850H with Radeon Graphics, 3201 Mhz, 8 Core(s), 16 Logical Processor(s)
- Tested locally on (2) Windows mini PC systems connected to LG OLED displays.
- Observed improved Dolby Vision vs HDR behavior during playback compared to current main branch behavior.


gnattu:
If I understand it correctly it does not do direct passthrough as that is ultimately an mpv limitation.

But for HDR playback handling part we need a bit more work because this is not a windows only issue.

For exmaple `target-colorspace-hint` is also needed for macOS (#293)

`hdr-compute-peak` is also not safe for all video inputs, as some of the videos has very weird metadata and enable this blindly will cause the tone mapping appear to be very weird.

I *think* we should keep those settings as custom mpv flags (if they can be set in that way now), or at least make them a client side settings so that the user can turn those off.

hausmanj:
Thanks, that makes sense.

The main goal of this PR was validating and preserving the Dolby Vision state through the JS/native/mpv playback pipeline so DV content could be handled differently from standard HDR10 content during playback.

My current Windows testing showed that conditional handling around `target-colorspace-hint` appeared to prevent the active playback fallback from DV to HDR on my AMD + LG setup, while `hdr-compute-peak=no` also improved behavior for my test movies. That said, I agree `hdr-compute-peak` is probably too aggressive as a blanket change given inconsistent metadata across sources.

Before revising further, I’m going to do more isolated testing to separate:

- DV detection/plumbing behavior
- `target-colorspace-hint` behavior
- `hdr-compute-peak` behavior

In addition, I am also going to test DV output on my M4 Pro to determine if this a broader mpv playback handling issue.

Based on what I find I’ll revise the PR toward a cleaner implementation or configurable behavior instead of broad global defaults.

hausmanj:
Updated the PR to narrow the scope further after additional isolated Windows testing.

Changes:

* Removed the `hdr-compute-peak=no` behavior entirely
* Kept the Dolby Vision metadata plumbing intact
* Restricted `target-colorspace-hint` handling to `_WIN32` only

Current Windows behavior:

* Dolby Vision titles correctly remain in DV during playback
* HDR10 titles correctly remain HDR
* Mixed HDR/DV transitions are stable (as tested on 2 client PCs)
* Theme/background video transitions are also behaving correctly after additional testing

At this point the effective playback behavior change is limited to Windows-only handling around `target-colorspace-hint` for detected Dolby Vision content.

hausmanj:
Additional update after further testing:

I also tested the current narrowed implementation on macOS (M4 Pro MacBook Pro connected to an LG OLED display).

Since the actual playback behavior change is now limited to the _WIN32 block around target-colorspace-hint, I mainly wanted to verify that the shared Dolby Vision metadata changes were not causing any side effects on macOS.

Results:
- HDR playback was still correctly detected by the LG TV
- I did not observe the dim/dull HDR behavior described in some of the other HDR issue reports
- Playback behavior appeared normal during my testing

At this point the PR behavior remains:
- shared DV metadata detection
- Windows-only mpv colorspace handling changes for detected Dolby Vision playback

andrewrabert:
Is this something mpv could handle internally? I'm hesitant here as I'm not sure.

Any issues in upstream mpv that relate?

gnattu:
> I did not observe the dim/dull HDR behavior described in some of the other HDR issue reports

Use your MacBook Pro internal display and it will be obvious. The HDR mode of the internal display will not even be triggered. But that is a separate issue.

hausmanj:
I was just searching around on the mpv upstream issues list... there are several loosely related discussions around target-colorspace-hint, HDR and Dolby Vision handling, but nothing that exactly matched this behavior.  So it seems like others have had challenges with the way mpv handles HDR and DV.

Thanks - I'll test with the internal display on my Mac - you're right I was looking at a 4k tv to compare.

hausmanj:
Update - I also ran into issue #387 when testing DV playback on the newer Rust builds, so I migrated this fix to the current Rust playback path.

Tested locally on (3) different Windows 11 Pro machines and it works well. The change is still isolated to Windows only.

I performed some additional validation:
- `cargo fmt --check` passes after formatting the Rust DV block.
- I also attempted `cargo clippy --workspace --all-targets -- -D warnings`, but it currently fails on unrelated upstream/tooling issues:
   - `jfn_rust/build.rs` clippy warning around `splitn`
   - `jfn-logging` Windows type alias acronym warnings
   - local `pkg-config` lookup during the clippy path

I did not make any changes for those clippy failures. The final branch only contains the migrated DV fix.

hausmanj:
After some additional testing and research I found a much cleaner way to handle this that removes all the frontend changes entirely.  The old approach passed a Dolby Vision flag from the JavaScript player through the native bridge into the Rust playback layer.  It worked but I sensed the concern with the comments provided.

The new approach just asks mpv directly. It turns out mpv already knows whether a file contains Dolby Vision — its demuxer reads that information from the video file and exposes it through a property called track-list. So instead of wiring up a signal from JS, we just query mpv after the file loads and read dolby-vision-profile straight from the track data. If it's there, we set target-colorspace-hint=no to preserve the DV signal; if not, we set yes for normal HDR.

The fix is now in a single file - api.rs. And it is still gated as Windows-only.

Turns out your earlier question about whether mpv could handle this internally was correct — it already exposes the data.

Tested and working on Windows across 3 machines. Cargo fmt and clippy both clean.

andrewrabert:
@hausmanj 

Oh nice! Makes me wonder if it could be a mpv Lua script instead. Any chance you could try to make one? I had luck pointing Claude at the mpv source+docs to create something similar in spirit #476

If you do, I'll look into setting up bundled Lua scripts ASAP so I can finally get this fix merged.

hausmanj:
@andrewrabert I quickly tested a draft Lua script and it works fine.  I'll keep testing it but so far so good.

Attached is what I used. (.lua instead of .txt)

[dv-detect.txt](https://github.com/user-attachments/files/29232358/dv-detect.txt)

hausmanj:
Updated — cross-platform Lua script bundling now complete (with help from claude)
  
  The PR has been expanded to include the full bundling infrastructure on all three platforms, plus additional scripts.

  How script delivery works:
  Scripts live in resources/mpv/scripts/ in the repo. At build time, xtask stages them next to the binary on Windows and
  Linux, and into Contents/Resources/mpv/scripts/ inside the .app bundle on macOS. At runtime, boot.rs scans that
  directory for .lua files, sorts them, and passes the list to mpv_set_option_string("scripts", ...) before
  mpv_initialize(). The path separator is ; on Windows and : on macOS and Linux. Each script is platform-guarded so it
  only activates where appropriate.

  Scripts bundled:
  dv-detect.lua — Windows only. Detects DV via track-list, sets target-colorspace-hint accordingly.

  macos-hdr-hint.lua — macOS only. Sets target-colorspace-hint=yes for EDR swapchain.

  linux-hdr-hint.lua — Linux only. Sets target-colorspace-hint=yes for Wayland compositor HDR surface.

  monoupmix.lua — All platforms. (Saw this in #403). Duplicates mono audio to L+R via pan filter; leaves stereo and  surround untouched. Credit to ctismer for the script.

Note: the xtask meson build now passes -Dlua=luajit explicitly on all platforms. On Windows, the mpv build
  previously had -Dlua=disabled in build_mpv_source.ps1 — that was flipped to luajit during development testing to
  validate the Lua script approach and remains in the PR as a required dependency. luajit will also need to be present in
  the build environment and CI dependencies for macOS and Linux builds.

Beta-Blaze:
blocking on this for dynamic RTX VSR scaling on Windows

resources/mpv/scripts/linux-hdr-hint.lua
```diff
@@ -0,0 +1,12 @@
+-- linux-hdr-hint.lua
```
andrewrabert, line 1: This doesn't seem to be needed in my experience. Were you seeing otherwise?

src/mpv/src/boot.rs
```diff
@@ -163,6 +258,20 @@ fn apply_defaults(
...
+    // Disable mpv's own default scripts-directory auto-load. Without this,
+    // mpv additionally scans its own config dir's `scripts/` subfolder and
+    // loads anything found there on top of `load_bundled_scripts` below —
+    // if a stray copy of a bundled script (e.g. dv-detect.lua, manually
+    // placed there during earlier testing) exists on a given machine, it
+    // runs a second time under an auto-suffixed name (e.g. `dv_detect2`),
+    // double-writing target-colorspace-hint around FILE_LOADED and
+    // corrupting fragile dual-layer DV/HEVC decode ("PPS changed between
+    // slices"). Bundled scripts are loaded explicitly below regardless.
+    set("load-scripts", "no")?;
```
andrewrabert, line 270: This shouldn't be needed. This app uses it's own special mpv config dir (eg. `~/.config/jellium-desktop/mpv` on Linux

hausmanj:
Hi, wow, I go away for a bit and looks like some things have changed lol.  I have been dealing with some DV locking issues in Windows that I've been troubleshooting that impacted switching to non-DV content.  Unfortunately I pushed the updates without fully assessing all of the recent changes, so I'l clean those up and realign to the compositor updates as well.  As for your comments, you're correct.  

The Linux HDR hint was me adding it to match with the macOS script, not because I hit anything. target-colorspace-hint already defaults to auto and Wayland’s a supported context for that, so it’s redundant, and forcing yes just makes it fire when the display parameters aren’t known. Keeping the macOS one though - cocoa isn’t in auto’s supported list (Wayland/D3D11/winvk), so it never engages there and the EDR swapchain never gets requested.

load-scripts=no - good catch, MPV_HOME does point at the app’s own dir. The double-load I hit was a copy of dv-detect.lua I’d hand-placed there myself while testing, so not a real scenario. It was also overreaching in a way I’d missed: options set via the API before mpv_initialize get command-line priority, so it’d have overridden load-scripts from a user’s own mpv.conf too. The hazard itself was real (two instances double-writing the hint around FILE_LOADED corrupted dual-layer DV/HEVC decode - “PPS changed between slices”), so dv-detect.lua now just claims a marker in user-data/ and any second instance bails immediately. Contained to the correct script now.

PR #359, opened 2026-05-13
fix(arm64): support native Linux aarch64 builds

luciancj:
## Summary

Fixes several issues that prevented `just build` from working on Linux aarch64 (tested on Asahi Fedora, Apple Silicon). Pre-built packages (AppImage, Flatpak) are x86_64-only, so native build is the only path for ARM64 Linux users.

**Root cause: CEF CMake ARM64 detection (`cmake/FindCEF.cmake`)**

`FindCEF.cmake` already passed `-DPROJECT_ARCH=arm64` when building the CEF wrapper — but only for Windows `ARM64`. On Linux, `CMAKE_SYSTEM_PROCESSOR` is `aarch64`, not `arm64`, so the condition was never true. The CEF wrapper was then configured with `-m64 -march=x86-64` flags (x86-only) and failed to compile on the first file.

Fix: match `aarch64|arm64|ARM64` on all platforms.

**Runtime: `EXTERNAL_MPV_DIR` auto-detect silently used system mpv (`CMakeLists.txt`)**

The auto-detect block defaulted to `/opt/jellyfin-desktop/libmpv` if that path existed. System mpv lacks the custom `wayland-display` / `wayland-surface` properties that the bundled fork exposes, causing `Platform init failed` at startup. Removed the auto-detect; `EXTERNAL_MPV_DIR` is now empty by default (build from submodule).

**Runtime: missing `cmake --install` rule for bundled `libmpv.so.2` (`CMakeLists.txt`)**

The POST_BUILD step correctly copies `libmpv.so.2` to the build directory, but there was no `install()` rule for the non-`EXTERNAL_MPV_DIR` case. `cmake --install` left stale files in place. Added `install(FILES ...)` + symlink creation for the bundled mpv path.

**Build: `-Dpipewire=disabled` in mpv meson setup (`CMakeLists.txt`)**

Prevents segfaults on systems where `libpipewire` is present (the fork's pipewire integration is not required since audio goes through alsa/pulseaudio).

**Fedora build deps (`dev/linux/install-deps-fedora.sh`)**

`plasma-wayland-protocols` and `xcb-util-devel` were not documented and caused build failures. Added a complete Fedora/RHEL dependency script with `just install-deps-fedora`.

## Test plan

- [x] `just build` succeeds on Linux x86_64 (existing CI)
- [x] `just build` succeeds on Linux aarch64 (Asahi Fedora, Apple M-series)
- [x] App launches and reaches Jellyfin web UI on Wayland/aarch64
- [x] `cmake --install` correctly places `libmpv.so.2` + `libmpv.so` symlink

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -443,15 +443,7 @@ endif()
 # mpv configuration
 option(BUILD_MPV_CLI "Also build the standalone mpv CLI binary from the submodule (for rendering comparisons; requires building mpv from submodule, not EXTERNAL_MPV_DIR)" OFF)
 
-# Auto-detect external mpv from /opt if it exists
-set(_DEFAULT_EXTERNAL_MPV_DIR "")
-if(EXISTS "/opt/jellyfin-desktop/libmpv/lib/libmpv.so")
-    set(_DEFAULT_EXTERNAL_MPV_DIR "/opt/jellyfin-desktop/libmpv")
-elseif(APPLE AND EXISTS "/opt/jellyfin-desktop/libmpv/lib/libmpv.dylib")
-    set(_DEFAULT_EXTERNAL_MPV_DIR "/opt/jellyfin-desktop/libmpv")
-endif()
-
-set(EXTERNAL_MPV_DIR "${_DEFAULT_EXTERNAL_MPV_DIR}" CACHE PATH "Path to external mpv installation (must contain include/ and lib/)")
+set(EXTERNAL_MPV_DIR "" CACHE PATH "Path to external mpv installation (must contain include/ and lib/); leave empty to build from third_party/mpv submodule")
 
 if(EXTERNAL_MPV_DIR)
     message(STATUS "Using external mpv from: ${EXTERNAL_MPV_DIR}")
@@ -490,7 +482,7 @@ else()
     if(NOT EXISTS "${MPV_BUILD_DIR}/build.ninja")
         message(STATUS "Configuring mpv with meson (cplayer=${_MPV_CPLAYER})...")
         execute_process(
-            COMMAND meson setup build --default-library=shared -Dlibmpv=true -Dcplayer=${_MPV_CPLAYER}
+            COMMAND meson setup build --default-library=shared -Dlibmpv=true -Dcplayer=${_MPV_CPLAYER} -Dpipewire=disabled
             WORKING_DIRECTORY ${MPV_SOURCE_DIR}
             RESULT_VARIABLE MPV_SETUP_RESULT
         )
@@ -885,6 +877,9 @@ else()
     install(DIRECTORY "${CEF_RESOURCE_DIR}/locales" DESTINATION .)
     if(EXTERNAL_MPV_DIR)
         install(FILES "${EXTERNAL_MPV_DIR}/lib/libmpv.so" DESTINATION .)
+    else()
+        install(FILES "${MPV_BUILD_DIR}/libmpv.so.2" DESTINATION .)
+        install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink libmpv.so.2 \"\${CMAKE_INSTALL_PREFIX}/libmpv.so\")")
     endif()
 endif()
 
--- a/README.md
+++ b/README.md
@@ -31,14 +31,30 @@ This project uses [just](https://github.com/casey/just) as a command runner.
 
 ```
 Available recipes:
-    appimage          # Build AppImage (outputs to dist/)
-    build             # Configure (if needed) + build the main app
-    clean             # Remove build artifacts (keeps CEF SDK download)
-    deps              # Ensure submodules and CEF are present
-    flatpak           # Build Flatpak bundle (outputs to dist/)
-    list              # List available recipes
-    mpv *args         # Run the standalone mpv CLI built from the submodule (forwards args)
-    run *args         # Run the app with debug logging (logs to build/run.log)
-    test              # Run unit tests
-    update-deps *args # Update vendored/fetched deps (CEF, doctest, quill); pass --check to verify only
+    appimage               # Build AppImage (outputs to dist/)
+    build                  # Configure (if needed) + build the main app
+    clean                  # Remove build artifacts (keeps CEF SDK download)
+    deps                   # Ensure submodules and CEF are present
+    flatpak                # Build Flatpak bundle (outputs to dist/)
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -443,15 +443,7 @@ endif()
 # mpv configuration
 option(BUILD_MPV_CLI "Also build the standalone mpv CLI binary from the submodule (for rendering comparisons; requires building mpv from submodule, not EXTERNAL_MPV_DIR)" OFF)
 
-# Auto-detect external mpv from /opt if it exists
```
andrewrabert, line 446: Remove this.

It's unrelated to any build problems you had.

CMakeLists.txt
```diff
@@ -490,7 +482,7 @@ else()
     if(NOT EXISTS "${MPV_BUILD_DIR}/build.ninja")
         message(STATUS "Configuring mpv with meson (cplayer=${_MPV_CPLAYER})...")
         execute_process(
-            COMMAND meson setup build --default-library=shared -Dlibmpv=true -Dcplayer=${_MPV_CPLAYER}
+            COMMAND meson setup build --default-library=shared -Dlibmpv=true -Dcplayer=${_MPV_CPLAYER} -Dpipewire=disabled
```
andrewrabert, line 485: Remove this.

If this is needed, I ask you open an issue first, then a dedicated PR. Universally disabling pipewire is not somehing I can make an informed decision on in the scoe of this PR.

dev/linux/install-deps-fedora.sh
```diff
@@ -0,0 +1,69 @@
...
+sudo dnf install -y \
+    cmake \
+    ninja-build \
+    gcc \
+    gcc-c++ \
+    git \
+    meson \
+    just \
+    python3 \
+    python3-pip \
```
andrewrabert, line 23: This shouldn't be needed.

CMakeLists.txt
```diff
@@ -885,6 +877,9 @@ else()
     install(DIRECTORY "${CEF_RESOURCE_DIR}/locales" DESTINATION .)
     if(EXTERNAL_MPV_DIR)
         install(FILES "${EXTERNAL_MPV_DIR}/lib/libmpv.so" DESTINATION .)
+    else()
+        install(FILES "${MPV_BUILD_DIR}/libmpv.so.2" DESTINATION .)
+        install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink libmpv.so.2 \"\${CMAKE_INSTALL_PREFIX}/libmpv.so\")")
```
andrewrabert, line 882: Why are these changes needed?

dev/linux/install-deps-fedora.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env sh
```
andrewrabert, line 1: Move this file to `dev/fedora/setup.sh`

The convention of `setup.sh` is already established for macOS (and similarly for windows with `setup.ps1`

dev/linux/install-deps-fedora.sh
```diff
@@ -0,0 +1,69 @@
...
+    libdvdnav-devel \
+    libcdio-paranoia-devel \
+    libbluray-devel \
+    libdisplay-info-devel \
+    libunwind-devel \
+    openal-soft-devel \
+    pulseaudio-libs-devel \
+    libsamplerate-devel \
+    openssl-devel \
+    libarchive-devel
```
andrewrabert, line 69: Assume mpv is being built with pipewire enabled. Add whatever pipewire `-dev` packages are needed for that.

dev/linux/linux.just
```diff
@@ -1,5 +1,10 @@
 # Linux-specific recipes (imported by top-level justfile)
 
+# Install build dependencies on Fedora/RHEL
+[linux]
+install-deps-fedora:
```
andrewrabert, line 5: I don't have a clear answer on how exactly should be available via just, but this is not it. Too distro-specific.

It also doesn't follow the convention of `just setup` like macOS and Windows have. The reason for this consistency is so that running `just run` works from a fresh clone on all the platforms. 

Remove any justfile modifications from this PR for now. I'll be adding Arch Linux support separately after I figure out how distro-detection should work.

CMakeLists.txt
andrewrabert, line 882: Why is this needed?

README.md
```diff
@@ -31,14 +31,30 @@ This project uses [just](https://github.com/casey/just) as a command runner.
...
+    flatpak                # Build Flatpak bundle (outputs to dist/)
+    install-deps-fedora    # Install build dependencies on Fedora/RHEL
+    list                   # List available recipes
+    mpv *args              # Run the standalone mpv CLI built from the submodule (forwards args)
+    run *args              # Run the app with debug logging (logs to build/run.log)
+    test                   # Run unit tests
+    update-deps *args      # Update vendored/fetched deps (CEF, doctest, quill); pass --check to verify only
 ```
+
+### Building from source on Linux
```
andrewrabert, line 47: Remove all changes to the README. Relates to https://github.com/jellyfin/jellyfin-desktop/pull/359/changes#r3237804409

andrewrabert:
This PR touches far more than it needs to - especially for a single commit. The PR summary is a bit lacking of reasoning for why to many of the changes. See feedback. 

Closing, but I welcome more targeted PR's in the future.

andrewrabert (changes requested):

cmake/FindCEF.cmake
```diff
@@ -133,7 +133,7 @@ elseif(CEF_ROOT AND EXISTS "${CEF_ROOT}/include/cef_version.h")
         if(CMAKE_GENERATOR)
             list(APPEND _CEF_CMAKE_ARGS -G "${CMAKE_GENERATOR}")
         endif()
-        if(WIN32 AND CMAKE_SYSTEM_PROCESSOR STREQUAL "ARM64")
+        if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|arm64|ARM64")
```
andrewrabert, line 136: Why lowercase `arm64`?

andrewrabert:
Native aarch64 AppImage builds were added in https://github.com/jellyfin/jellyfin-desktop/pull/361

Additionally, it's also possible to efficiently develop locally without needing to install deps. 
```
just appimage build
just appimage run
```

PR #360, opened 2026-05-13
windows: fix mouse and context menu offset on HiDPI displays

ChristianEmilsson:
## Summary

Fixes mouse click and context menu positioning being offset on Windows HiDPI displays (e.g. 4K at 150% scaling). The offset scales with the DPI factor and is most noticeable on high-resolution monitors in multi-monitor setups.

**Root cause:** `win_get_scale()` relies on mpv's `display-hidpi-scale`, which reports `1.0` in per-monitor DPI-aware mode on Windows even when display scaling is active. This causes `logical == physical` dimensions everywhere, so:
- Mouse coordinates are sent to CEF in physical pixels instead of logical
- `GetViewRect` / `GetScreenInfo` report incorrect dimensions and `device_scale_factor`
- Context menu `GetXCoord()`/`GetYCoord()` are in paint-buffer space but positioned in CSS pixels

**Fix:**
- `win_get_scale()` now queries `GetDpiForWindow()` directly instead of trusting mpv's value, falling back to mpv / `GetDpiForSystem()` when unavailable
- Mouse coordinates in `input_windows.cpp` are converted from physical to logical before dispatching to CEF
- Context menu coordinates in `RunContextMenu` are divided by `device_scale_factor` before passing to JavaScript

## Test plan

- [ ] On a Windows HiDPI display (e.g. 4K at 150%), verify mouse clicks land on the correct UI elements
- [ ] Right-click on a video and verify the context menu appears at the cursor position
- [ ] Drag the window between monitors with different DPI and verify both mouse and context menu stay correct
- [ ] Verify no regression on a 1x (96 DPI) display

**Environment:** Windows 11, 4K + HD dual-monitor setup, 150% scaling on 4K monitor

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cef/cef_client.cpp
+++ b/src/cef/cef_client.cpp
@@ -763,10 +763,19 @@ bool CefLayer::RunContextMenu(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame>
     if (pending_menu_callback_) pending_menu_callback_->Cancel();
     pending_menu_callback_ = callback;
 
+    // GetXCoord/GetYCoord return coordinates in the paint-buffer (physical)
+    // space, but the JS menu uses CSS position:fixed in logical/CSS pixels.
+    // Divide by device_scale_factor to convert.
+    float scale = (physical_w_ > 0 && width_ > 0)
+        ? static_cast<float>(physical_w_) / width_
+        : 1.0f;
+    int cx = static_cast<int>(params->GetXCoord() / scale);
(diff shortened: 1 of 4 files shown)


andrewrabert:
I'm not sure this will get merged, but I am actively working through similar issues that intermittently occur on Wayland. Looking to land it as a unified solution.

andrewrabert:
Still working on this. Will post a PR when ready.

PR #365, opened 2026-05-16
Rust

andrewrabert:
This PR changes a number of things to Rust. The immediate motivation behind this was introducing https://github.com/mahkoh/wl-proxy as a way to solve a number of Wayland size/scale issues without further modifying mpv.


src/single_instance/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+        addr.sun_family = AF_UNIX as _;
+        let bytes = path.to_bytes_with_nul();
+        if bytes.len() > addr.sun_path.len() {
+            return None;
+        }
+        for (dst, src) in addr.sun_path.iter_mut().zip(bytes.iter()) {
+            *dst = *src as c_char;
+        }
+        Some(addr)
+    }
```
CounterPillow, line 83: lmfao. "The original used strncpy, better reimplement an ad-hoc string copy in unsafe rust, this will surely be more maintainable and safer for my code!"

PR #383, opened 2026-05-18
windows: fix build tooling for PowerShell and VS detection

### LINKED ISSUE #382: Windows: build tooling broken on PowerShell 5.1 / VS detection fails
### Confirmation

- [x] I am running a supported version as per the [Supported Version](https://github.com/jellyfin/jellyfin-desktop/wiki/Supported-Versions) guide.

### Version

3.0.0-dev (6feeaca)

### Platform

Windows

### OS version

_No response_

### GPU

_No response_

### Describe the bug

 Description: The Windows build scripts have several issues that prevent just deps and just build from working in a standard PowerShell session:

  1. just has no shell on Windows — just defaults to sh, which isn't in PATH in a normal PowerShell session. Without set windows-shell, every recipe fails immediately.
  2. build_mpv_source.ps1 uses PS 7 syntax but runs under PS 5.1 — setup.ps1 invokes the script with powershell.exe (PS 5.1), but the script contains syntax only valid in PS 7 (&& in strings, etc.), causing a cascade of parse errors.
  3. vswhere picks up the wrong Visual Studio install — without the -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 filter, vswhere -products * can match other Microsoft products (e.g. SQL Server Management Studio) that don't ship VC++ tools, causing vcvars64.bat lookup to fail.
  4. vcvars loading broken in PS 7 — cmd /c ""$VcVars" && set" fails in PowerShell 7 due to changed native command argument passing behavior, so MSVC environment variables are never imported.
  5. gendef not found — the pacman package list was missing mingw-w64-clang-x86_64-tools, which is the package that provides gendef in the CLANG64 environment.
  6. set positional-arguments in justfile breaks script invocations — this causes the recipe name to be appended as an extra argument to every script call, breaking [ValidateSet] parameters in PowerShell scripts.

### Steps to reproduce

  1. Fresh Windows machine with Visual Studio Build Tools, MSYS2, and just installed
  2. Clone the repo
  3. Run just deps from a standard PowerShell terminal
  4. Observe failures at various stages depending on environment

  Expected: just deps and just build complete successfully

  Actual: Multiple failures — no shell found, PS 5.1 parse errors, MSVC not detected, gendef not found

### Logs

_No response_

### Additional context

_No response_

jburgoy03:
- windows.just: add windows-shell directive so just uses cmd.exe instead of sh; invoke build.ps1 via pwsh to avoid PS 5.1 parse errors
 - justfile: remove set positional-arguments (appended recipe name as an extra argument to pwsh scripts, breaking ValidateSet parameters)
 - setup.ps1: install and use pwsh (PS 7) to run build_mpv_source.ps1
 - build_mpv_source.ps1: filter vswhere to VC.Tools to avoid matching non-VS products; use temp batch file to load vcvars (PS 7 native command arg passing broke the inline cmd /c approach); add mingw-w64-clang-x86_64-tools package (provides gendef)
 - build.ps1: same vswhere filter and temp-bat vcvars fixes
  
Fixes #382

### DIFF
--- a/dev/windows/build.ps1
+++ b/dev/windows/build.ps1
@@ -19,14 +19,18 @@ if (-not $env:VSINSTALLDIR) {
 
     $VsWhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
     if (Test-Path $VsWhere) {
-        $VsPath = & $VsWhere -latest -products * -property installationPath
+        $VsPath = & $VsWhere -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
         $VcVars = Join-Path $VsPath "VC\Auxiliary\Build\vcvars64.bat"
         if (Test-Path $VcVars) {
-            cmd /c "`"$VcVars`" && set" | ForEach-Object {
+            $TempBat = Join-Path $env:TEMP "jfn_vcvars_build.bat"
+            Set-Content $TempBat -Value ('@call "' + $VcVars + '"') -Encoding ASCII
+            Add-Content $TempBat -Value '@set' -Encoding ASCII
+            cmd /c $TempBat | ForEach-Object {
                 if ($_ -match "^([^=]+)=(.*)$") {
                     [Environment]::SetEnvironmentVariable($matches[1], $matches[2], "Process")
                 }
             }
+            Remove-Item $TempBat -ErrorAction SilentlyContinue
             Write-Host "Loaded Visual Studio environment" -ForegroundColor Green
         }
     }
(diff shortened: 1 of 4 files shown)


andrewrabert:
Sorry about the recent Windows failures stemming from the Rust conversion.

I did get it working on Windows again in https://github.com/jellyfin/jellyfin-desktop/pull/379 though probably in a hackish way. 

Lmk if there's anything you'd like to change there and I'll merge.

jburgoy03:
> Sorry about the recent Windows failures stemming from the Rust conversion.
> 
> I did get it working on Windows again in #379 though probably in a hackish way.
> 
> Lmk if there's anything you'd like to change there and I'll merge.

Yeah no worries I noticed the shutdown issue which I assumed was an easy-ish fix and then ran into some startup struggles which this resolves and hopefully eases some pain points if anyone else is running Windows

PR #393, opened 2026-05-20
perf(settings): coalesce save_async into a single persistent worker

ctismer:
`jfn_settings_save_async()` spawned a detached thread per call. Concurrent saves fought the same file mutex, and any in-flight write could be killed mid-flush on shutdown.

This change adds a single persistent worker (`Condvar` + `Mutex<Pending>`), started lazily on first `save_async`. Queued writes coalesce — only the newest snapshot survives. New `jfn_settings_shutdown_save_worker()` drains the pending write and joins the thread; called from `main.cpp` right after the synchronous final save so nothing is lost when CEF teardown begins.

Rust port of `9a215d9` (which targeted the old C++ Settings class that no longer exists after the settings → Rust port).

Smoke-tested locally on macOS 12 Monterey x86_64: Cmd-Q after window move correctly persists window geometry to `settings.json`.

### DIFF
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -26,7 +26,12 @@ void jfn_settings_init(const char* path);
 
 bool jfn_settings_load(void);
 bool jfn_settings_save(void);
+// Fire-and-forget save: hands a snapshot to a single persistent background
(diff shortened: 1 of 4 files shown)


andrewrabert:
Thanks!

PR #397, opened 2026-05-21
playback: make per-item position survive sigkill

ctismer:
The throttled HTTP position reports can lose the last sample on a hard exit, and the next launch rewinds by minutes.

This adds a per-item position cache at <config-dir>/playback-position.json, written at ~1 Hz with tmp+rename+fsync so the last sample survives kill -9.

On LoadStarting a fresh entry (>= 5 s, within 7 days) is emitted as SeekAbsolute after FileLoaded. EOF clears it. Cap 256 entries.

Tested on macOS 12 x86_64.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -285,6 +285,8 @@ static int run_with_cef(int mw, int mh,
     // Coordinator + sinks must exist before any thread can post inputs or
     // observe playback state. Sinks register before start() so the worker
     // never delivers to a half-built fanout.
+    jfn_playback_position_init(
+        (paths::getConfigDir() + "/playback-position.json").c_str());
     PlaybackCoordinatorScope coord_scope;
     // Builtin idle_inhibit sink (Rust-side) calls into the platform vtable
     // through this handler. Install before any event posts.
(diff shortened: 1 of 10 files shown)


andrewrabert:
I'm torn on this one and am hesitant due to it breaking the server as the source of truth.

I wonder if this could be made as an mpv LUA plugin instead?
- https://github.com/jellyfin/jellyfin-desktop/wiki/MPV-Configuration
- https://mpv.io/manual/stable/#lua-scripting

ctismer:
The real reason I added this brutal workaround is a different problem, at least on my old MacBook on Monterey: when I leave Jellyfin Desktop open (playback stopped) and let the display sleep, then wake it with a keystroke or mouse move, the app window comes back frozen. Force-restarting then loses the last playback position.

So what actually needs addressing is the freeze.

rodis120:
I also experienced app freezing after leaving video paused for some but in my case it happened on Wayland.

However, I think that saving video progress locally would cause problems for people who watch video on multiple devices, as it will continue playing video from locally saved timestamp instead of the moment in which a user stopped watching on different device.

ctismer:
@rodis120 But would not the latest client which continued the movie do the update, which is correct?
Anyway, fixing the freeze problem instead.

ctismer:
Filed #414 to fix the underlying wake-from-sleep freeze directly. That removes the original motivation for this workaround — this can be closed in favor of #414.

PR #403, opened 2026-05-22
Default audio-channels to stereo

ctismer:
Without this change, mono audio sources play out of the left speaker only on macOS. The default mpv behavior is to pass mono straight through to the audio sink and CoreAudio routes the single channel to the left output only — so a mono mkv (e.g. an Oceans 11 1960 rip) is silent on the right. Tested on Monterey x86_64 against a Frank Sinatra mono mkv: before the patch, right channel silent; after, both speakers carry the mono content.

The fix is a single `set("audio-channels", "stereo")` in `apply_defaults`. Users who explicitly configure another layout (5.1, 7.1, auto) via the existing Audio Channel Layout setting still get their choice — `apply_boot_options` runs after `apply_defaults` and overrides.

### DIFF
--- a/src/mpv/src/boot.rs
+++ b/src/mpv/src/boot.rs
@@ -125,6 +125,13 @@ fn apply_defaults(
     // by language / default-flag / codec scoring.
     set("track-auto-selection", "no")?;
 
+    // Force stereo output by default so mono sources are duplicated to
+    // both speakers instead of left-only (the default mpv behavior is to
+    // pass mono straight through, which CoreAudio on macOS routes to the
+    // left channel only). Users with surround setups can override via the
+    // Audio Channel Layout setting; that override runs in apply_boot_options.
+    set("audio-channels", "stereo")?;
+
     // Input: we own all devices and route through CEF.
     set("input-default-bindings", "no")?;
     set("input-vo-keyboard", "no")?;
--- a/src/mpv/src/boot.rs
+++ b/src/mpv/src/boot.rs
@@ -121,6 +121,13 @@ fn apply_defaults(handle: &Handle, display: DisplayBackend) -> crate::error::Res
     // by language / default-flag / codec scoring.
     set("track-auto-selection", "no")?;
 
+    // Force stereo output by default so mono sources are duplicated to
+    // both speakers instead of left-only (the default mpv behavior is to
+    // pass mono straight through, which CoreAudio on macOS routes to the
+    // left channel only). Users with surround setups can override via the
+    // Audio Channel Layout setting; that override runs in apply_boot_options.
+    set("audio-channels", "stereo")?;
+
     // Input: we own all devices and route through CEF.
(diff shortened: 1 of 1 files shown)


andrewrabert:
Still thinking about this one. Leaving a note for my futureself:
https://mpv.io/manual/stable/#options-audio-channels

ctismer:
It came up when we viewed "Ocean's Eleven" from 1960 on Jellyfin. This is a mono movie, and the sound came only out of the left channel. We needed to go back using VLC which played it correctly.

andrewrabert:
I'm curious if this could be solved with an mpv lua script that upmixes all mono audio tracks to the maximum number of channels on the playback audio device. If possible, that would eliminate the need entirely to change the preferred output configuration.

A mono upmix script is something I'd be open to bundling and enabling by default. The infrastructure to bundle isn't there yet, but will be added for the sake of https://github.com/jellyfin/jellyfin-desktop/pull/346

ctismer:
Agreed — a mono-only upmix is the right shape; a global stereo default downmixes surround for everyone, which is too broad.

I wrote a small mpv Lua script doing exactly what you described: on file-load and on audio-track switch it reads the selected track's source channel count and, only when it's mono, appends a tagged `pan` filter duplicating the single channel onto L/R. Multichannel tracks are left completely untouched.

Verified locally on macOS with standalone mpv: a mono source opens the AO as stereo with both channels carrying identical content (per-channel RMS equal at -21.2 dB); stereo and 5.1 test files pass through unchanged, no filter added.

```lua
local mp = require 'mp'
local msg = require 'mp.msg'

local TAG = "monoupmix"
local applied = false

-- Source channel count of the currently selected audio track (pre-filter).
local function selected_source_channels()
    local n = mp.get_property_number("track-list/count", 0)
    for i = 0, n - 1 do
        local t = string.format("track-list/%d/", i)
        if mp.get_property(t .. "type") == "audio"
            and mp.get_property_bool(t .. "selected", false) then
            return mp.get_property_number(t .. "demux-channel-count")
        end
    end
    return nil
end

local function update()
    local mono = selected_source_channels() == 1
    if mono and not applied then
        mp.commandv("no-osd", "af", "append",
            "@" .. TAG .. ":lavfi=[pan=stereo|c0=c0|c1=c0]")
        applied = true
        msg.info("mono source detected -> duplicating to stereo")
    elseif not mono and applied then
        mp.commandv("no-osd", "af", "remove", "@" .. TAG)
        applied = false
    end
end

mp.register_event("file-loaded", update)
mp.observe_property("current-tracks/audio/id", "native", update)
```

Happy to target the device's full layout instead of stereo if you prefer, and to land it wherever the bundling infra from #346 ends up. Shall I convert this PR into the script, or keep the stereo default as a stopgap until bundling exists? Either is fine by me.

*(Script drafted with AI assistance, verified locally.)*

PR #405, opened 2026-05-23
Add config option to ignore SSL errors

### LINKED ISSUE #82: Allow configuration for Self-Signed SSL certificates
**Describe the feature**
Add a config option that allows connecting to a URL with self-signed / invalid SSL certificates.

**Did Jellyfin Media Player / the Qt desktop client have this feature?**
Yes

**Additional context**
Used to be configurable with `ignoreSSLErrors=true` in the json config in the Qt client.

BMagnu:
Implements and closes #82.
Pretty much just reintroduces a catch-all flag like the old client.
Internally, it just passes a similar ignore-ssl-error flag to CEF and lets it handle that automatically.


andrewrabert:
I'll look into this in the near future after a little more cleanup up the Rust rewrite.

PR #409, opened 2026-05-23
Implement overlay fading as css animation

rodis120:
Platform specific code for a fading animation that is only visible for split of a second at the start felt to me like a overkill, so I implemented it in css.

It also fixes problem with lack of support for `wp_alpha_modifier_v1` protocol on some Wayland compositors.


andrewrabert:
I'll look into this in the near future after a little more cleanup up the Rust rewrite.

rodis120:
Let me know when you finish. I will rebase it after that.

andrewrabert:
Nice code reduction. 

I merged your commit + a separate one-line fade-in fix via https://github.com/jellyfin/jellyfin-desktop/pull/422

PR #414, opened 2026-05-24
macos: reload browsers on wake from sleep

ctismer:
After display sleep, CEF subprocesses retain a stale GPU context and the
renderer stops dispatching input. The GUI stays on the last frame but
does not respond to clicks; the only recovery is force-quit and restart.

Subscribe to NSWorkspaceDidWakeNotification and reload every CEF browser
on fire. The wake observer is macOS-only; the reload primitive on the
layer is cross-platform.

Tested manually on MBP-2015 (Intel Iris Pro, Monterey): display sleep,
idle wake, click resumes responsiveness instantly.

### DIFF
--- a/src/jfn_cef/src/browsers.rs
+++ b/src/jfn_cef/src/browsers.rs
@@ -290,6 +290,21 @@ pub extern "C" fn jfn_browsers_send_external_begin_frame_all() {
     }
 }
 
+/// Reload every layer's browser. Used by platform wake-from-sleep handlers
+/// to recover from stale GPU context / IPC after the system was asleep.
+#[unsafe(no_mangle)]
+pub extern "C" fn jfn_browsers_reload_all() {
+    let snapshot: Vec<*mut JfnCefLayer> = INSTANCE
+        .lock()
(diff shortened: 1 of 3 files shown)


andrewrabert:
Curious - does https://github.com/jellyfin/jellyfin-desktop/pull/420 also solve the issue for you?

I wasn't able to replicate the issue, but I'm hoping it fixes rendering without reloading the content. 

If it doesn't, could you share some logs from when you saw the stuck GUI issue?

ctismer:
#420 fixes it for me. Repro on Intel Mac (Iris Pro, Monterey): `pmset displaysleepnow` then wake — with #414 the UI stayed frozen, with #420 it repaints immediately. Log at wake:

```
[WAKE] wake/screen-change; restarting display link
[CVDL] restarted
```

The display-only-sleep path (ScreensDidWake) is what bit me — #414 only observed DidWake so it missed that case, and even when it did fire, reloading the page content didn't revive the stranded CADisplayLink. Closing #414 in favor of this.

PR #446, opened 2026-05-29
Fix X11 select dropdown popups - Issue 253

hausmanj:
While working on X11, I remembered the dropdown issue and looked into how the dropdowns are displayed.

It looks like the X11 side had placeholder code for showing dropdown lists, so the app knew a dropdown should open but nothing was actually being shown. 

This adds the missing X11 handling so dropdown lists can appear, update, close etc.

Tested on a Linux/X11 session using Parallels on my Mac and they seem to work now.

Adding additional validation completed - ran cargo fmt and ran cargo clippy -p jfn-x11 --all-targets -- -D warnings and all good.

### DIFF
--- a/src/x11/src/lifecycle.rs
+++ b/src/x11/src/lifecycle.rs
@@ -3,6 +3,8 @@
 
 use std::sync::Arc;
 
+use crate::shm::shm_free;
+
 use xcb::{Xid, XidNew, x};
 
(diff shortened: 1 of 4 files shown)


andrewrabert:
X11 had decent refactor. Handled dropdowns a bit differently in https://github.com/jellyfin/jellyfin-desktop/pull/468

PR #475, opened 2026-06-08
Reload mpv.conf on playerLoad

fatmalama:
Invoke (new) `jfn_mpv_reload_config_file()` from `handle_player_load` which calls mpv's `load-config-file` with `mpv_home()`/mpv.conf as arg.

Applies changes from mpv.conf when (re-)starting playback, without having to reopen jellyfin.

In case there's no mpv.conf, mpv logs `[mpv] file: Cannot open file <mpv.conf path>: No such file or directory`, but other than that does not crash and does not negatively impact playback.

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -2285,6 +2285,7 @@ version = "3.0.0-dev"
 dependencies = [
  "bindgen",
  "jfn-color",
+ "jfn-paths",
  "libc",
  "objc2",
  "parking_lot",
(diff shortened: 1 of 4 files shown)


andrewrabert:
I'm not sure how this would look, but the implementation here isn't what I'd go with. I'd try to find a LUA based approach that does this so Jellyfin Desktop code doesn't even know about it. A hook trigger is where I'd start looking. https://mpv.io/manual/stable/#hooks

Let's flesh it out in the issue first before another PR on this.

fatmalama:
@andrewrabert alright, I've no idea about mpv's lua integration, but in that case #476 can stay open while someone manages to make such trigger

fatmalama:
~actually I'm not even finding any lua stuff on the repo other than on third_party/mpv which is mpv's source code. where should I be looking for?~ continuing conversation on #476

PR #503, opened 2026-06-15
Wayland: release speculative popup grab on button release

### LINKED ISSUE #494: On Hyprland, I cannot click on other windows while Jellyfin Desktop is opened
### Confirmation

- [x] I am running a supported version as per the [Supported Version](https://github.com/jellyfin/jellyfin-desktop/wiki/Supported-Versions) guide.

### Version

JellyfinDesktop-3.0.0-dev+7de39a1-x86_64.AppImage

### Platform

Linux / Wayland

### OS version

EndeavourOS/Arch Linux 7.0.11

### GPU

Nvidia 4090

### Describe the bug

I'm not able to click in other windows while Jellyfin is opened after moving workspaces for some reason.

Using hyprland 0.55.4

### Steps to reproduce

1. open Jellyfin Desktop 
2. Try to click into my other window
3. Clicks register in Jellyfin desktop rather than the window I selected.

### Logs

_No response_

### Additional context

_No response_

Its-Haze:
popup::arm() requests an xdg_popup.grab on every left/right pointer press
so the grab carries that press's still-live serial; the menu model only
arrives later via CEF's async callback, and the popup is left unmapped
until then.

On Mutter/KWin an unmapped popup's grab is inert, so a press that never
opens a menu is harmless. On wlroots/Hyprland the grab takes effect the
instant it is requested, even while the popup is unmapped: a menu-less
press then holds a seat grab that captures all input until the next press
happens to break it — and the last press before the pointer goes idle is
never broken, freezing input on the window entirely. Clicking elsewhere
cannot escape because the grab routes those clicks back to this surface,
which re-arms the grab.

Release the speculatively-armed grab popup on the triggering button's
release when no menu model has arrived. This runs between the press and
the next one, so it never races a subsequent press's fresh arm (unlike
the CEF-driven hide() path), and it leaves real menus untouched — they
set a model before release, so the grab they need is preserved.

Fixes #494

### DIFF
--- a/src/wayland/src/input.rs
+++ b/src/wayland/src/input.rs
@@ -399,6 +399,10 @@ impl Dispatch<wl_pointer::WlPointer, ()> for State {
                         state.cef_modifiers(),
                     );
                 }
+                // Drop the grab armed on the press if this click opened no menu (#494).
+                if (button == BTN_RIGHT || button == BTN_LEFT) && !pressed {
+                    crate::popup::dismiss_if_speculative();
+                }
             }
             Event::Axis { axis, value, .. } => {
                 if matches!(axis, WEnum::Value(wl_pointer::Axis::VerticalScroll)) {
--- a/src/wayland/src/popup.rs
+++ b/src/wayland/src/popup.rs
@@ -361,6 +361,24 @@ pub(crate) fn on_done(generation: u32) {
     clear_menu_locked(&mut st);
 }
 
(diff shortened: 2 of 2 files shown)


Meisterlala:
I build this locally and tested it.
This fixes the #494 for me on hyperland

andrewrabert:
Thanks!

PR #527, opened 2026-06-22
Better packaging

xaltsc:
Closes #525

Blocked by #515

PR #528, opened 2026-06-22
Add CI support

### LINKED ISSUE #526: flake: provide scripts to automate flake updating
The PR #527 repackages `cef-binary` to keep it up to date with the version of CEF used the `cef` crate (although it is not used if `nixpkgs` is up to date, but this repo moves faster than `nixpkgs` which has had a PR https://github.com/NixOS/nixpkgs/pull/522930 updating `cef-binary` for ages). This is currently manually done, which is less than ideal.

Something should be added to CI to automate this task.

Some inspiration can maybe be drawn from here https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/ce/cef-binary/update.sh

Another thing that should be automated is updating flake inputs (i.e. `nix flake update`).

xaltsc:
I'm trying to address the issues in #526.

So far, I've got a script that updates nix's `cef-binary` to match the version used by the rust crate `cef`.

Blocked by #527

Closes #526


xaltsc:
Please verify the files `.github/workflows/cef-nix-sync.yml` and `renovate.json`.

This PR is untested btw, I just know the `_update-cef.sh` script works on my machine; I have no idea how to test it otherwise.

xaltsc:
~For the record: the standard ways (i.e. `Mic92/nix-update` and the like) of updating the updated `cef-binary` package don't work because I couldn't find a way to override the `filename` attribute this derivation contains (which still points to the original path in `nixpkgs`).~

This is fixed #527

xaltsc:
There's also a way to run `nix flake update` using GitHub workflows instead of references, cf https://github.com/xaltsc/jellyfin-desktop/blob/main/.github/workflows/update-flake-lock.yml .

I don't know which one is preferable.

xaltsc:
So, besides rebasing on the latest version of #527, I've improved the workflows and added one that updates the flake. The two workflows are tested (except for the hook on `renovate`'s PRs about updating CEF) as you can see on my own fork https://github.com/xaltsc/jellyfin-desktop (although I made them before editing this PR).

There are two things that are supposed to update the flake's inputs: the workflow in the last commit, which is tested, and the `renovate` thing, which ... exists. You probably don't want both, I'm not able to decide which is best, so do tell me which one you want and I'll remove one of the commits.

Except for the points mentioned above, I consider this PR complete.

xaltsc:
So I added this workflow https://github.com/xaltsc/jellyfin-desktop/blob/flake-ci/.github/workflows/cef-nix-sync-in-pr.yml which should update nix's cef-binary within the same PR as Renovate's ones instead of in another.

This makes sense because, in the meantime between the update of rust `cef` and that of nix `cef-binary` the package is broken.

Again, I have no idea what I'm doing, so quintuple check what I've done.

Also, this makes the other workflow obsolete, but, thus far, I've kept it as a working example before review.

xaltsc:
I removed the old workflow that opened a PR to update `nix` cef. The new one is much better and works, cf: https://github.com/xaltsc/jellyfin-desktop/pull/6
https://github.com/xaltsc/jellyfin-desktop/actions/runs/28466217566

Issue #47, opened 2026-03-28
Add settings option to make subtitles bold


Pentaphon:
**Describe the feature**
Please add an option to change between normal and bold subtitles.

**Did Jellyfin Media Player / the Qt desktop client have this feature?**
No 

**Describe alternatives you've considered**
Or possibly just make the client respect the server's subtitle weight setting?

Issue #534, opened 2026-06-23
macos: clicking on dropdown items doesn't commit the selected option


Sheeplet1:
### Confirmation

- [x] I am running a supported version as per the [Supported Version](https://github.com/jellyfin/jellyfin-desktop/wiki/Supported-Versions) guide.
- [x] I acknowledge that the version of Jellyfin Desktop I'm reporting for starts with `3.0.0-dev`
- [x] I will disclose any use of custom styles or plugins that alter the web appearance.

### Version

3.0.0-dev+04fae1c

### Platform

macOS

### OS version

MacOS 27

### Describe the bug

The Mac implementation for dropdowns uses `NSMenu` but mouse selections does not commit the selected value back to CEF. 

1. CEF opens the internal `<select>` pop-up.
2. Jellyfin Desktop shows `NSMenu` on top of it.
3. Clicking a menu item (`NSMenuItem`) does not update the DOM `<select>`.
4. The current implementation seems to replay arrow-key events plus Enter into the internal select rather than setting the DOM value directly. 

To be clear, clicking menu items doesn't work. If the select has focus, then keyboard input works and we can use arrow keys to select and commit the value.

### Steps to reproduce

1. Go into `Settings` and any sub-setting like `Playback`.
2. Try to select an item from a dropdown menu.

### Additional context

https://github.com/user-attachments/assets/b1741641-12f6-426d-9dcb-3111476d8ed4

PR #535, opened 2026-06-23
config: add subtitle font size scaling and font weight

### LINKED ISSUE #182: Add option for subtitle size
**Describe the feature**
Subtitle size setting in the menu

**Did Jellyfin Media player have it 
Yes

Sheeplet1:
## Description

This resolves #182 and #47 by introducing two playback settings: subtitle size and subtitle bold. 

They're exposed through `windows.jmpInfo.settings.playback` and `settingDescriptions.playback` so they'll show up in the app's playback settings UI.

### Implementation

1. We persist these settings in `jfn_config`:
- `subtitleBold` defaults to `false`.
- `subtitleScale` defaults to `1.0`
- `subtitleScale` is validated against set values: `[0.5, 0.75, 1.0, 1.25, 1.5]`. 

The subtitle scale options mirrors the five scaling options from the old Jellyfin player.

2. CEF handles `subtitleBold` and `subtitleScale` setting updates
3. `jfn_mpv::api` exposes accessors for `sub-scale` and `sub-bold` so saved settings are applied during playback.

### Screenshots

<details>
  <summary>Screenshots</summary>

  #### Settings UI

Screenshot 2026-06-30 at 7 56 04 pm

  #### Set to **Default**

  Screenshot 2026-06-23 at 9 22 07 pm

  #### Set to **1.5** and **Bold**

  Screenshot 2026-06-23 at 9 23 44 pm

</details>

### DIFF
--- a/src/web/native-shim.js
+++ b/src/web/native-shim.js
@@ -91,7 +91,9 @@
         settings: {
             main: { enableMPV: true, fullscreen: false, userWebClient: '__SERVER_URL__' },
             playback: {
-                hwdec: _savedSettings.hwdec || 'auto'
+                hwdec: _savedSettings.hwdec || 'auto',
+                subtitleBold: !!_savedSettings.subtitleBold,
+                subtitleScale: _savedSettings.subtitleScale ?? 1.0
             },
             audio: {
                 audioPassthrough: _savedSettings.audioPassthrough || '',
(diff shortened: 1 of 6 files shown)


Sheeplet1:
Also found a bug with the dropdown menus on Mac, reported in #534

src/web/native-shim.js
```diff
@@ -111,7 +113,15 @@
         },
         settingsDescriptions: {
             playback: [
-                { key: 'hwdec', displayName: 'Hardware Decoding', help: 'Hardware video decoding mode. Use "auto" for automatic detection or "no" to disable.', options: _savedSettings.hwdecOptions }
+                { key: 'hwdec', displayName: 'Hardware Decoding', help: 'Hardware video decoding mode. Use "auto" for automatic detection or "no" to disable.', options: _savedSettings.hwdecOptions },
+                { key: 'subtitleScale', displayName: 'Subtitle Size', options: [
+                    { value: 0.5, title: 'Tiny' },
```
andrewrabert: Ditch the made-up mapping and just show the scale factor directly. 

The ideal state would allow text input for any factor, but I don't wanna block the PR on that.

andrewrabert (changes requested):

src/web/native-shim.js
Sheeplet1: @andrewrabert What do you think of a slider and text input? mpv's subscale accepts from 0 to 100

Sheeplet1: It would look something like this 

Screenshot 2026-06-24 at 7 50 15 pm

Issue #539, opened 2026-06-24
Flake: provide devshells


xaltsc:
### Confirmation

- [x] I have confirmed this feature is absent from a [supported version](https://github.com/jellyfin/jellyfin-desktop/wiki/Supported-Versions).

### Version

any

### Describe the feature

Nix's devshells provide an environment for development, like Python's `venv`.

It would be nice to have one for this project.

@Sheeplet1 has already done some work here #538, but it's incompatible with my other `nix` related PRs.

Again, as in the other issues/PRs, I'm far from an expert in any field I use for this, I just know how to use a search engine and read. So comments, suggestions and critiques are very much welcome.

PR #538, opened 2026-06-24
Setup local dev environment

Sheeplet1:
## Description

Adds a Nix-based local dev environment. 

- This has only been tested and built on NixOS. I plan to test and refine on MacOS later. 
- The flake layout follows how Ghostty's done it.

## Screenshots

image

PR #543, opened 2026-06-24
Add devshells

xaltsc:
This is an attempt at solving #539.

The current commit is an extremely crude approximation of what is needed. **PLEASE DO NOT MERGE**.

There's a competing effort in PR #538.

Blocked by #528

### DIFF
--- a/.github/workflows/cef-nix-sync.yml
+++ b/.github/workflows/cef-nix-sync.yml
@@ -1,0 +1,55 @@
+name: CEF version nix sync
+
+on:
+  push:
+    branches:
+      - "**"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  regen:
+    runs-on: ubuntu-latest
+    # Only fire when the triggering (head) commit is a Renovate cef bump.
+    if: startsWith(github.event.head_commit.message, 'Update Rust crate cef')
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v6
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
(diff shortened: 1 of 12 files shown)


xaltsc:
So, I'm trying to use `nixpkgs''s `mpv` libs as requested by @andrewrabert.

`$JFN_MPV_INCLUDE_DIR` has been set to various values, however, it's still impossible to find `client.h`.

I'm getting the following error when running `just clippy`:

```
  --- stderr
  Error: "could not locate mpv/client.h in any of: [\"third_party/mpv/include\"]\nSet JFN_MPV_INCLUDE_DIR, EXTERNAL_MPV_DIR, or install libmpv via pkg-config."
```

So it doesn't look like `JFN_MPV_INCLUDE_DIR` is considered at all.

xaltsc:
Also, when running `just clippy` with `CEF_PATH` set (to the wrong one, but that's not the issue), I get

```
  CEF_PATH is invalid (File I/O error: failed to open file `/nix/store/v4lvgffqpc359ya581qw0p6fcralbpfk-cef-binary-149.0.4/archive.json`: No such file or directory (os error 2)),
```

No such `archive.json` exist. I assume it's something that exist in the downloaded tarball which is the scenario that `clippy` expects. But build works without it.

xaltsc:
Also, `just build` fetches `mpv` and CEF, which is unnecessary, and even tries to build them (afaict, `libplacebo` is not a dependency of `jfnd` itself).

xaltsc:
I don't think I can work on this PR further without taking a maximalist approach (which I refuse to). 

Build methods needs to be patched before I can advance.

andrewrabert:

Issue #522, opened 2026-06-21
No window on Niri


xaltsc:
### Confirmation

- [x] I am running a supported version as per the [Supported Version](https://github.com/jellyfin/jellyfin-desktop/wiki/Supported-Versions) guide.
- [x] I acknowledge that the version of Jellyfin Desktop I'm reporting for starts with `3.0.0-dev`
- [x] I will disclose any use of custom styles or plugins that alter the web appearance.

### Version

3.0.0-various

The last commit from the main repo for the nix package is 096a01257fc4993d1b3eb785c5d8176f734d9364.

`flatpak run org.jellyfin.JellyfinDesktop -v` returns 
```
jellyfin-desktop 3.0.0-dev+676919e

CEF 149.0.4+g2f1bfd8+chromium-149.0.7827.156

mpv-version mpv v0.41.0-UNKNOWN
ffmpeg-version 7.1.3
```

### Platform

Linux / Wayland

### OS version

NixOS Unstable with Niri as the compositor

### GPU

8-core AMD Ryzen 7 8845HS w/ Radeon 780M Graphics (-MT MCP-)

### Describe the bug

This is two related bugs. On both `flatpak` and `nix` (cf. #515), `jfnd` fails to present a working window.

The executable built by `nix` completely fails to present a window. More details are available in the pull request #515 as I tried to investigate the issue before realising that the executable worked on Gnome.

For `flatpak`, a blank window is shown:

<img width="1172" height="2064" alt="Image" src="https://github.com/user-attachments/assets/30dd1408-6427-45db-8473-43222a351d6d" />

Below is the output of `flatpak run org.jellyfin.JellyfinDesktop --log-level=debug 2>jfnd.log`.

```
F: /run/user/1000/.flatpak/org.jellyfin.JellyfinDesktop/xdg-run/wayland-1 is not a symlink to "../../flatpak/wayland-1" as expected: readlinkat: L’argument passat no és vàlid
[Main] jellyfin-desktop 3.0.0-dev+676919e
[Main] CEF 149.0.4+g2f1bfd8+chromium-149.0.7827.156
[Main] Display backend: wayland
[Main] wlproxy listening on wayland-1
[CEF] wlproxy: State::build: could not create a socket: Too many open files (os error 24)
[CEF] wlproxy: dispatch: the server hung up the connection
[CEF] wlproxy: dispatch: the server hung up the connection
[CEF] wlproxy: dispatch: the server hung up the connection
[...]

PR #554, opened 2026-06-26
Compile with `crane`

### LINKED ISSUE #545: Flake: use `crane` as an alternative build tool
The current PR #513 uses `nixpkgs` default rust tools because the ultimate goal is to publish on `nixpkgs` and that's the only thing acceptable.

The cost of this is that it rebuilds everything everytime something in the derivation slightly changes.

There exists `crane` https://github.com/ipetkov/crane which is much more tightly woven into the rust ecosystem, and in particular, builds cargo dependencies as separate derivations, thus saving resources.

While the current derivation is to be kept for the reasons mentioned above, `crane` should probably be used as the default mechanism to build `jfnd`.

xaltsc:
Closes #545 

Blocked by #527 

See the later comments in #527 for some more context.

~This PR is WIP and not ready for merging.~


xaltsc:
This PR seems ready for merging. I'm still a bit unsure of how crane works and would have to experiment with crate updates, but it looks like it does what it's supposed to do.

Build works, but I couldn't test it due to #522. Built against #563 with success.

andrewrabert:

PR #563, opened 2026-07-01
Fully demote mpv to Wayland subsurface

andrewrabert:
tldr on Wayland:
- **Before:** mpv created the Wayland window and Jellyfin Desktop hijacked it's Wayland connection to create layers on top of it (with hacks relating to window resize & state)
- **After:** Jellyfin Desktop creates the Wayland window and connection. Mpv is embedded solely as a subsurface of that window. Jellyfin Desktop controls window resize & state management exclusively.

---

Previously, mpv established the Wayland connection with jfnd tapping into it with wl-proxy. Less of a proxy and more of an intercept.

This PR fully encapsulates mpv's still-established socket by wrapping with a second wl-proxy that sits as a clean edge between pure-mpv and the rest of jnfd's Wayland communication. This allows jfnd to atomically commit reszes of all subsurfaces in a single transaction.

Could use cleanup, but deferring to later.

Fixes:
https://github.com/jellyfin/jellyfin-desktop/issues/522
https://github.com/jellyfin/jellyfin-desktop/issues/588

LLMs were used in the creation of this PR.


src/wayland/src/mpv_proxy.rs
```diff
@@ -0,0 +1,1019 @@
...
+    WindowSize::new(CUR_W.load(Ordering::Acquire), CUR_H.load(Ordering::Acquire))
+}
+
+/// Owns the S_mpv thread's wake eventfd. Writers signal, and teardown clears it,
+/// under this lock — so a wake can never write to a descriptor that has been
+/// closed (and possibly reused) out from under it.
+static MPV_WAKE: Mutex<Option<WakeEvent>> = Mutex::new(None);
+
+pub fn set_window_size(w: c_int, h: c_int) {
+    if let Some(size) = WindowSize::new(w, h) {
```
MBR-0001, line 95: this is allegedly a torn read/write, my source may be incorrect though, seems pretty minor in any case

src/wayland/src/wl_state.rs
```diff
@@ -245,18 +306,115 @@ noop_dispatch!(
...
+            mgd.swap_remove(pos);
+            buf.buf.destroy();
+        }
+        Some(pos) => mgd[pos].doomed = Some(buf.buf),
+        None => {
+            debug_assert!(
+                false,
+                "retire_buffer: untracked buffer — a release may have been missed"
+            );
+            tracing::error!("retire_buffer: untracked buffer — a release may have been missed");
```
MBR-0001, line 366: [minor] claude claims this may be a memory leak, suggests `buf.buf.destroy();` instead of mgd.push
```
The recovery code says: "I don't recognize this buffer, so I'll hold onto it and free it
once the compositor tells me it's safe."

But the reason it doesn't recognize the buffer is that the connection to the compositor's
signals is already broken for it — so "wait for the signal" means "wait forever."
```

MBR-0001 (commented):
I let claude review this and went through its findings, since I don't know rust I'm leaving this as "comment" instead of request changes

my input may be totally useless but at least I tried

src/wayland/src/mpv_proxy.rs
andrewrabert: Will be addressed in a future pass.

src/wayland/src/wl_state.rs
andrewrabert: Will be addressed in a future pass.

PR #532, opened 2026-06-22
Add home-manager module

xaltsc:
This PR is tested with success.

Blocked by #527

Closes #531

PR #566, opened 2026-07-01
add flake cache

### LINKED ISSUE #542: Flake: set up a cache
### Confirmation

- [x] I have confirmed this feature is absent from a [supported version](https://github.com/jellyfin/jellyfin-desktop/wiki/Supported-Versions).

### Version

any

### Describe the feature

While providing recipes for `nix` users to build the package from source is nice, rebuilding it on every machine is a waste of resources.

Nix's outputs (for instance, packages) can be stored in caches. The only third-party one I know of is `cachix`, which is widely used, and which I thought was free for FOSS project, but I cannot find the info anywhere.

Customer service for `cachix` says that it is indeed free for open-source projects, but they couldn't provide a source.

### Additional context

_No response_

xaltsc:
Closes #542 

Blocked by #554 #532 #528 

There's some infrastructure to set up, namely, a cachix.org cache and setting secret variables to set in the repo; which I could take care of, if I'm given the permission.

Regarding the cache, if the `nix` effort is to be extended to other Jellyfin projects (I'm mostly thinking of the server and `jellyfin-mpv-shim`), it might be a good idea to simply name it `jellyfin` and use it as well for these repos.

Some values are currently placeholder values, given that there's no official cache set up.

It looks like it's working perfectly https://github.com/xaltsc/jellyfin-desktop/actions/runs/28536851231/job/84600309929 ; sadly, I can't really test it given that all the package on my system are up to date.

Before merging this PR, a lot of rebasing will be needed. Only the two latest commits matter.


xaltsc:
I added a runner to cache builds for `aarch64` which seems to work https://github.com/xaltsc/jellyfin-desktop/actions/runs/28628135036/job/84899080861 .

I haven't been able to test the result in a VM though.

andrewrabert:

PR #580, opened 2026-07-03
Feature/user profile switching

hausmanj:
Summary: One thing missing has been easy profile switching in the app, similar to other popular streaming apps.  This PR adds fast switching between saved user profiles on the same server, so you can move between accounts without signing in again. 

A Select User item in preferences under the User section opens a "Who's watching?" picker. You can also add a new user from there and back out of the login screen if you change your mind.

It's implemented as an injected web script (src/web/user-switch.js) wired through the existing embedded-JS injection pipeline, with a small amount of native focus handling.

This also includes the fixes that were needed to run the client on Linux during development. The most important one fixes a black screen on load under Parallels: on libmpv builds without Wayland support (Ubuntu 24.04, wayland-client < 1.23) the gpu-context=wayland option doesn't exist, so setting it aborted startup — now it falls back to X11/EGL via XWayland. The others keep injected scripts alive across a browser destroy and filter control characters out of Wayland keyboard events.

Note: If only one profile is logged in the Select User Screen does not appear, but the Select User setting is still there where you can add users.

Tested on Windows 25H.2, Mac 26.6, X11, Wayland and Fedora

image

image

image

### DIFF
--- a/src/jfn_cef/src/app.rs
+++ b/src/jfn_cef/src/app.rs
@@ -191,10 +191,22 @@ wrap_render_process_handler! {
                 .insert(id, ExtraInfo::from_dictionary(extra.clone()));
         }
 
-        fn on_browser_destroyed(&self, browser: Option<&mut Browser>) {
-            let Some(browser) = browser else { return };
-            let id = browser.identifier();
-            self.inner.profiles.lock().remove(&id);
+        fn on_browser_destroyed(&self, _browser: Option<&mut Browser>) {
+            // Deliberately not removing the profile from `self.inner.profiles`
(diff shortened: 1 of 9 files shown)


hausmanj:
I realized during testing that there are two additional User Settings areas, one at the bottom of the hamburger menu on the left side, and the Dashboard avatar menu on the top right.  I've now added Select User settings in both of those as well - way more user friendly since its easier to find.

image

Hamburger Menu

image

Dashboard avatar menu

Re-tested across Mac, Windows, Fedora, Ubuntu X11 and Wayland

PR #581, opened 2026-07-06
wlproxy: fall back to wl_shm when single-pixel-buffer is missing

bogdan-velicu:
## The problem

On compositors that don't implement `wp_single_pixel_buffer_manager_v1` — notably niri (YaLTeR/niri#619) — the app starts, connects to the server, and loads the full UI, but **no window ever appears**.

`wlproxy` uses single-pixel-buffer unconditionally for the two solid-color buffers the root window needs (the root background in `fill_root()` and the transparent overlay backdrop in `adopt_host_surface()`). When the global is missing, both functions bail early, no buffer is ever attached to the root surface, and an xdg_toplevel without a buffer is never mapped — so the app runs headless. The only diagnostic is the `wlproxy: missing globals for root` stderr line.

## The fix

**Commit 1:** route all solid-color buffers through a new `create_rgba_buffer()` helper:
- if the compositor advertises single-pixel-buffer, behavior is unchanged;
- otherwise build the equivalent 1×1 ARGB8888 buffer via `wl_shm` (memfd + 4-byte pool), which every compositor supports. The viewporter scaling path is identical either way.

The globals check becomes `single_pixel_buffer || wl_shm`.

**Commit 2 (related, separable):** set `title`/`app_id` on the root toplevel. mpv's own `set_title` lands on its demoted toplevel, so the real window is currently nameless on *every* compositor — no taskbar label, no icon association, and compositor window rules can't match it. The app id matches the desktop file (`org.jellyfin.JellyfinDesktop`). Happy to split this into its own PR if preferred.

## Testing

- niri 26.04 (no single-pixel-buffer): window now maps and renders the full UI; title and app id show correctly. Before the patch: headless.
- Verified the registry dump on niri lacks `wp_single_pixel_buffer_manager_v1` but has `wl_shm`, `wl_compositor`, `wl_subcompositor`, `xdg_wm_base`, `wp_viewporter`.
- `cargo fmt --check` and `cargo clippy` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/wlproxy/src/lib.rs
+++ b/src/wlproxy/src/lib.rs
@@ -1315,6 +1315,10 @@ fn build_root() {
     let root_tl = root_xdg.create_child::<XdgToplevel>();
     root_tl.set_handler(RootToplevelH);
     root_xdg.send_get_toplevel(&root_tl);
+    // mpv's own set_title/set_app_id land on its demoted toplevel, so the
+    // root would otherwise stay nameless (breaks taskbars and window rules).
+    root_tl.send_set_title("Jellyfin Desktop");
+    root_tl.send_set_app_id("org.jellyfin.JellyfinDesktop");
 
     if INITIAL_MAXIMIZED.load(Ordering::Acquire) {
         root_tl.send_set_maximized();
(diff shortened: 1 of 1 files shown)


andrewrabert:
Should be addressed already in https://github.com/jellyfin/jellyfin-desktop/pull/563

PR #592, opened 2026-07-11
Add localized desktop settings and menus

nx111:
The desktop shim and native CEF menus currently contain hard-coded English labels, while Jellyfin Web and the rest of the application are localized. This makes the settings and application menus inconsistent for non-English users.

Add a small CEF-side translation resource, inject the selected desktop locale into the web shim, and use localized labels for settings sections, options, menus, and about information. Keep English fallbacks for missing keys and preserve the existing native/Web locale behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,3 +56,11 @@ Available recipes:
     run *args     # Run the app
     run-mpv *args # Run the mpv CLI
 ```
+
+### Desktop Localization
+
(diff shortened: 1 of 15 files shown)


andrewrabert:
Thanks for this. Going to close as I'm rewriting the entire Jellium Desktop client configuration+about UI in Iced. 

Not sure when that will be ready for localization to be introduced

Iced client UI rewrite: https://github.com/andrewrabert/jellium-desktop/pull/669
Issue tracking localization support: https://github.com/andrewrabert/jellium-desktop/issues/673

PR #605, opened 2026-07-16
fix: Fix compilation on latest nightly

Jacoblightning:
`fetch_update` is [deprecated in 1.99.0](https://doc.rust-lang.org/std/sync/atomic/type.AtomicU32.html#method.fetch_update) and was renamed to `try_update`.

This PR updates it to the new name to fix compilation on nightly/any version above 1.99.0

All tests pass on both stable and nightly with this change.


Jacoblightning:
No longer needed.
Jellium desktop now compiles on nightly since [39787bf](https://github.com/andrewrabert/jellium-desktop/commit/39787bf41d4f3016422b5e2acb2e909288e1d8e7)

PR #617, opened 2026-07-18
wayland: implement client-side key repeat

mysticalsoap:
## Problem
Wayland has no protocol-level auto-repeat like X11 — the compositor sends `wl_keyboard::Event::RepeatInfo` (rate/delay) once and expects the client to resynthesize keydowns itself while a key is held. The `Dispatch<WlKeyboard>` impl only forwarded raw press/release events and dropped `RepeatInfo` entirely, so holding a key (arrow keys in a list/grid, letters in a text field) fired a single keydown with no repeat.

## Fix
Adds a Linux timerfd to `State`, armed on a repeatable keydown (checked via `xkb::Keymap::key_repeats`) using the compositor-reported rate/delay (clamped so a reported delay/rate of 0 can't make `timerfd_settime` disarm the timer entirely), and polled alongside the existing display/wake fds in `worker_loop`. Each expiration resends the held key through the same `cb.key`/`cb.char_` path a real keydown uses. The timer is disarmed on key release — including a release that's swallowed while a native popup (context menu / `<select>`) is active, so a stale repeat can't keep firing after the popup closes — on real focus loss (alt-tab/workspace switch), and proactively the moment a right-click or Menu-key press might open a popup, rather than waiting on the popup's own (asynchronous) active state. X11/macOS/Windows already get repeat from the OS and are unaffected — this is Wayland-only.

## Verification
`just fmt`/`just lint`/`just build` all pass. Manually confirmed: holding an arrow key scrolls a media list/grid continuously; holding a letter repeats in a text field; repeat stops on release and on alt-tab away mid-hold; repeat doesn't leak into an open context menu or `<select>` dropdown; releasing a held key while a popup is open and then closing the popup doesn't leave it stuck auto-repeating.

Assisted-by: claude-sonnet-5

### DIFF
--- a/src/wayland/src/input.rs
+++ b/src/wayland/src/input.rs
@@ -8,7 +8,7 @@
 
 use parking_lot::Mutex;
 use std::ffi::{c_int, c_void};
-use std::os::fd::{AsFd, AsRawFd};
+use std::os::fd::{AsFd, AsRawFd, FromRawFd, OwnedFd};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 use std::thread::{self, JoinHandle};
@@ -158,6 +158,11 @@ struct State {
     cursor_type: Arc<AtomicU32>,
 
     menu_focus: bool,
+
+    repeat_timer_fd: OwnedFd,
+    repeat_rate: i32,
+    repeat_delay: i32,
+    repeat_key: Option<u32>,
 }
 
 impl State {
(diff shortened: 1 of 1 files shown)


mysticalsoap:
hello,

baby first vibe coded PR. tried to keep it reigned in on comments/in general. sorry for the bother, this issue is just a real annoyance on my tv/wireless keyboard setup and this app works great otherwise for me. primarily sonnet but used fable to double check changes.

andrewrabert:
This is great, thank you!

PR #593, opened 2026-07-11
Fix Windows HiDPI maximize and CEF viewport sizing

nx111:
## Problem

On Windows 11 with PerMonitorV2 and fractional display scaling, especially at
250% on a 3840x2160 display, the desktop window could:

- Maximize with incorrect non-client geometry.
- Leave blank margins around the window or content.
- Keep a stale CEF viewport while the home page was idle.
- Crop or offset the lower/right part of the rendered content.
- Convert mouse coordinates incorrectly for the embedded input window.
- Remain stuck during startup while waiting for the mpv window.

Older profiles could also retain Jellyfin Web's `layout=tv` preference, which
hides the horizontal navigation buttons on media rows.

## Changes

- Embed a PerMonitorV2 DPI manifest.
- Keep the main, mpv, and input threads in a consistent DPI context.
- Use monitor-specific DPI and DPI-aware decorated frame calculations.
- Query the settled native client size during startup and native resize events.
- Propagate resize updates to CEF even when no media is playing.
- Round logical dimensions upward at fractional scales.
- Convert Windows input coordinates to the browser's logical coordinate space.
- Crop accelerated CEF frames to their visible rectangle before presentation.
- Migrate stale `layout=tv` preferences to `layout=desktop` once.

## Scope

The `third_party/mpv` submodule pointer is intentionally unchanged. The PR
does not depend on unpublished mpv commits, so GitHub Actions can fetch and
build the existing CI-pinned revision.

## Validation

- `cargo test --manifest-path src/Cargo.toml -p jfn-playback`
  - 46 tests passed.
- `cargo test --manifest-path src/Cargo.toml -p jfn-windows`
  - Passed; crate currently has no unit tests.
- `node --check src/web/native-shim.js`
- Windows release build completed successfully.
- Manually validated on Windows 11 with a 3840x2160 HDMI display at 250%
  scaling, including maximize/restore, home-page scrolling, and video playback.

PR #627, opened 2026-07-21
Fix Windows HiDPI scaling: declare PerMonitorV2 before mpv, map input to logical coords

### LINKED ISSUE #1: Program crashes due to failing to create Vulkan device
```
    ~  jellyfin-desktop-cef                                                                                                                                                        ✔ 
Calling CefExecuteProcess...
CefExecuteProcess returned: -1
[EGL] Initialized EGL 1.5
[EGL] DMA-BUF import: yes
[EGL] Context created successfully
[EGL] GL_VERSION: OpenGL ES 3.2 NVIDIA 580.119.02
[EGL] GL_RENDERER: NVIDIA GeForce RTX 4070 SUPER/PCIe/SSE2
Bound wp_color_manager_v1
Created mpv subsurface below main window
Created color management surface
Failed to create Vulkan device
Fatal: Wayland subsurface init failed
```

```
    ~  fastfetch                                                                                                                                                                   ✔ 
██████████████████  ████████    chris@chris
██████████████████  ████████    -----------
██████████████████  ████████    OS: Manjaro Linux x86_64
██████████████████  ████████    Kernel: Linux 6.12.63-1-MANJARO
████████            ████████    Uptime: 4 hours, 19 mins
████████  ████████  ████████    Packages: 1601 (pacman)[stable]
████████  ████████  ████████    Shell: zsh 5.9
████████  ████████  ████████    Display (S34C65xU): 3440x1440 in 34", 100 Hz [External]
████████  ████████  ████████    DE: KDE Plasma 6.5.4
████████  ████████  ████████    WM: KWin (Wayland)
████████  ████████  ████████    WM Theme: Breeze
████████  ████████  ████████    Theme: Breeze (BreathDark) [Qt], Breeze-Dark [GTK2], Breeze [GTK3]
████████  ████████  ████████    Icons: breeze [Qt], breeze [GTK2/3/4]
████████  ████████  ████████    Font: Noto Sans (10pt) [Qt], Noto Sans (10pt) [GTK2/3/4]
                                Cursor: breeze (24px)
                                Terminal: konsole 25.12.0
                                Terminal Font: Hack Nerd Font Mono (11pt)
                                CPU: Intel(R) Core(TM) i3-8350K (4) @ 4.00 GHz
                                GPU: NVIDIA GeForce RTX 4070 SUPER [Discrete]
                                Memory: 8.66 GiB / 31.26 GiB (28%)
                                Swap: Disabled
                                Disk (/): 2.11 TiB / 3.58 TiB (59%) - ext4
                                Disk (/data): 24.20 TiB / 39.39 TiB (61%) - ext4
                                Local IP (wlp5s0): 192.168.0.101/24
                                Locale: en_GB.UTF-8


```

```
    ~  nvidia-smi                                                                                                                                                                1 ✘ 
Fri Jan 16 14:00:34 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.119.02             Driver Version: 580.119.02     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 4070 ...    Off |   00000000:01:00.0  On |                  N/A |
|  0%   36C    P8              4W /  220W |    1109MiB /  12282MiB |      1%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
```

Ive also attached the output of `vulkaninfo`

### LINKED ISSUE #2: Gamescope overrides vk_xwayland_wait_ready and triggers RADV VRS errors
## Description

When running the app inside Gamescope on Bazzite (AMD GPU / RADV), the following errors appear in logs:

```
[CEF] ATTENTION: default value of option vk_xwayland_wait_ready overridden by environment.
[CEF] radv: Can't open file: '/tmp/radv_vrs.CXkMjzrv'.
[CEF] radv: Failed to initialize the notifier for RADV_FORCE_VRS_CONFIG_FILE!
```

This does **not** occur when running directly in KDE (Wayland session) outside of Gamescope.

I suspect this may cause proper hdr to fail:

KDE
```
[mpv] cplayer: VO: [libmpv] 3840x1600 yuv420p10
[mpv] libmpv_render: target colorspace: trc=12 prim=6 peak=0
```

VS

```
[mpv] [cplayer] VO: [libmpv] 3840x1600 p010
[mpv] [libmpv_render] target colorspace: trc=2 prim=3 peak=0
```

### LINKED ISSUE #625: Windows: maximised window uses incorrect size at 125% display scaling on a 4K monitor
### Confirmation

### Version

0.1.0-dev+79c0354

### Platform

Windows

### OS version

Windows 11

### GPU

NVIDIA RTX 3080

### Describe the bug

This is a vanilla Jellyfin Web UI with no custom CSS or UI altering plugins.

On Windows with a 4K display configured to use 125% scaling, Jellium Desktop does not initially fill the available screen when launched with its saved maximised state.

The application opens in the top-left of the screen and occupies only part of the display, despite appearing to be maximised.

If I manually un-maximise the window and maximise it again, Windows resizes it correctly and it fills the full available screen.

I have tried "windowMaximized": true, "windowMaximized": false in the settings.json and fullscreen=no, fullscreen=yes in the mvp.conf with no changes to the bugs behaviour, only changing Windows display scaling from 125% to 100% resolved the issue. At 100%, Jellium launches maximised at the correct size.

### Steps to reproduce

1. Connect or use a 3840 × 2160 display.
2. Set Windows display scaling to 125%.
3. Set the following in Jellium's settings.json:
"windowMaximized": true
4. mpv.conf contains:
fullscreen=no
5. Completely close Jellium Desktop.
6. Launch Jellium Desktop again.

### Logs

_No response_

### Additional context

- The problem appears to be DPI-scaling related.
- At 100% Windows scaling, Jellium launches maximised correctly.
- At 125% scaling, the initial maximise operation appears to use incorrect physical or logical dimensions.
- Manually restoring and remaximising triggers Windows to calculate the correct window size.

vuksv:
# Fix Windows HiDPI scaling: declare PerMonitorV2 before mpv, map input to logical coords

## Problem
On Windows with a scaled display (4K TVs/monitors at 125–300%), the Jellyfin
UI renders at the wrong scale — everything laid out tiny/half-size and packed
into part of the window — and mouse hit-testing is misaligned so buttons appear
not to work while scaled. At 100% everything is fine. Fixes #625.

## Root cause
The process never declares DPI awareness. The exe ships **no application
manifest** and nothing calls `SetProcessDpiAwareness*` at startup, so the
process launches **DPI-unaware**. The startup order is then:

1. mpv creates its window while the process is still DPI-unaware, so
   `GetDpiForWindow` returns **96** and mpv reports `display-hidpi-scale=1`.
2. `CefInitialize` runs afterwards and Chromium flips the whole process to
   **PerMonitorV2** — too late. The window now really reports its true DPI, but
   mpv has already latched scale 1.

Because scale is authoritatively sourced from mpv's `display-hidpi-scale`, the
CEF overlay is sized for a physical-pixel *logical* viewport (e.g. 3840 instead
of 1920 at 200%), so the web UI is laid out at the wrong scale. Mouse
coordinates are separately affected (see fix #2).

## Fix
Two small changes, both in `src/windows`:

1. **Declare Per-Monitor-V2 DPI awareness before mpv starts.** `win_early_init()`
   runs before the mpv proxy, so a single
   `SetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2)`
   there makes mpv detect the real monitor DPI and report the correct
   `display-hidpi-scale`. Using the API (not a manifest) keeps the change to one
   file and avoids adding a `<supportedOS>`/compatibility block; it requires
   Windows 10 1703+, which CEF already mandates, and silently no-ops on anything
   older (falling back to today's behaviour).

2. **Convert mouse input from physical to logical coordinates.** CEF's
   off-screen renderer expects mouse events in logical/DIP space (matching its
   view rect), but the Win32 input child window reports physical pixels. This
   only worked before because the broken scale of 1.0 made physical == logical;
   once fix #1 makes the scale correct, un-converted clicks would land at
   `scale`× the intended point. A small `phys_to_logical()` helper divides the
   incoming coordinates by the current display scale at every mouse dispatch
   site (move, buttons, both wheels).

## Relation to #593
Same root cause as #593. This is a much smaller change: it does not need the
CEF frame cropping, idle-resize propagation, upward rounding, cross-thread DPI
juggling, or the `layout=tv` migration from that PR. Establishing awareness
process-wide at startup makes the per-thread DPI-context work unnecessary, and
common scales (125/150/200/300%) divide the 3840-wide panel evenly so no
rounding compensation is required. No manifest is added, so there is no
pre-Windows-10 `<supportedOS>` entry to remove.

## Testing
- Built and run on a 3840×2160 display.
- Verified `display-hidpi-scale` now matches the OS scale (observed at 200% and
  300%); UI renders at correct size and fills the maximized window.
- Display scale can be changed live without relaunching — the app rescales
  correctly (PerMonitorV2 `WM_DPICHANGED` handled via mpv).
- `cargo fmt --check` and `cargo clippy -D warnings` pass.

## Screenshots

repro_aware

fixed_300

### DIFF
--- a/src/windows/src/input.rs
+++ b/src/windows/src/input.rs
@@ -279,6 +279,19 @@ fn is_button_down(msg: u32) -> bool {
     matches!(msg, WM_LBUTTONDOWN | WM_RBUTTONDOWN | WM_MBUTTONDOWN)
 }
 
+#[inline]
+fn phys_to_logical(x: c_int, y: c_int) -> (c_int, c_int) {
+    let scale = crate::platform::win_get_scale();
+    if scale > 0.0 && (scale - 1.0).abs() > f32::EPSILON {
+        (
+            (x as f32 / scale).round() as c_int,
+            (y as f32 / scale).round() as c_int,
+        )
+    } else {
+        (x, y)
+    }
+}
+
 // =====================================================================
 // WndProc.
 // =====================================================================
@@ -298,12 +311,8 @@ unsafe extern "system" fn input_wndproc(hwnd: HWND, msg: u32, wp: WPARAM, lp: LP
         }
 
         WM_MOUSEMOVE => {
-            jfn_input_dispatch_mouse_move(
-                get_x_lparam(lp),
-                get_y_lparam(lp),
-                mouse_modifiers(wp),
-                0,
-            );
+            let (lx, ly) = phys_to_logical(get_x_lparam(lp), get_y_lparam(lp));
+            jfn_input_dispatch_mouse_move(lx, ly, mouse_modifiers(wp), 0);
             return LRESULT(0);
         }
 
@@ -318,11 +327,12 @@ unsafe extern "system" fn input_wndproc(hwnd: HWND, msg: u32, wp: WPARAM, lp: LP
             if down {
                 let _ = unsafe { SetFocus(Some(hwnd)) };
             }
+            let (lx, ly) = phys_to_logical(get_x_lparam(lp), get_y_lparam(lp));
             jfn_input_dispatch_mouse_button(
                 msg_to_button_code(msg),
                 if down { 1 } else { 0 },
-                get_x_lparam(lp),
-                get_y_lparam(lp),
+                lx,
+                ly,
                 mouse_modifiers(wp),
             );
             return LRESULT(0);
@@ -358,8 +368,9 @@ unsafe extern "system" fn input_wndproc(hwnd: HWND, msg: u32, wp: WPARAM, lp: LP
             unsafe {
                 let _ = ScreenToClient(hwnd, &mut pt);
             }
+            let (lx, ly) = phys_to_logical(pt.x, pt.y);
             let delta = hiword_i16(wp.0 as u32) as i32;
-            jfn_input_dispatch_scroll(pt.x, pt.y, 0, delta, mouse_modifiers(wp));
+            jfn_input_dispatch_scroll(lx, ly, 0, delta, mouse_modifiers(wp));
             return LRESULT(0);
         }
 
@@ -371,8 +382,9 @@ unsafe extern "system" fn input_wndproc(hwnd: HWND, msg: u32, wp: WPARAM, lp: LP
             unsafe {
                 let _ = ScreenToClient(hwnd, &mut pt);
             }
+            let (lx, ly) = phys_to_logical(pt.x, pt.y);
             let delta = hiword_i16(wp.0 as u32) as i32;
-            jfn_input_dispatch_scroll(pt.x, pt.y, delta, 0, mouse_modifiers(wp));
+            jfn_input_dispatch_scroll(lx, ly, delta, 0, mouse_modifiers(wp));
             return LRESULT(0);
         }
 
--- a/src/windows/src/platform.rs
+++ b/src/windows/src/platform.rs
@@ -18,7 +18,9 @@ use windows::Win32::Graphics::Gdi::{
     GetMonitorInfoW, HMONITOR, MONITOR_DEFAULTTONEAREST, MONITORINFO, MonitorFromWindow,
 };
 use windows::Win32::UI::Controls::MARGINS;
-use windows::Win32::UI::HiDpi::GetDpiForSystem;
+use windows::Win32::UI::HiDpi::{
+    DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2, GetDpiForSystem, SetProcessDpiAwarenessContext,
+};
 use windows::Win32::UI::WindowsAndMessaging::{
     CWPRETSTRUCT, CallNextHookEx, GWL_STYLE, GetWindowLongPtrW, GetWindowRect,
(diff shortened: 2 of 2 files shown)


Uncle-Tio:
Tested this PR on Windows 11, 3440×1440 display at 125% scaling, fixes the maximize bug for me.

The startup log shows `display-hidpi-scale=1.25`, and the CEF layer syncs to `logical=2752x1080 physical=3440x1351`, the window fills the screen on launch and the web UI is laid out at the correct scale.

Issue #632, opened 2026-07-23
NVIDIA/Wayland: CEF shared textures blocked by implicit-modifier probe + upstream Chromium OSR bug


mysticalsoap:
### Version
0.1.0-dev+e9df081

### Platform
Linux / Wayland (KDE Plasma 6.7.3 / KWin)

### GPU
NVIDIA GeForce RTX 4090 (driver 610.43.03)

### Custom styles / plugins (disclosure)
Running the Abyss theme plus Media Bar Enhanced and Jellyfin Enhanced. The stock UI on the software fallback is tolerable; the heavier theme is what made the lag noticeable.

### Summary
This is a diagnosis rather than a plain bug report. NVIDIA/Wayland can't use CEF's shared-texture (GPU) rendering path, blocked by *two independent* problems: the first is fixable here, the second is an upstream Chromium bug that fixing the first only exposes. Net result either way — NVIDIA stays on `--disable-gpu-compositing` software CEF rasterization, as it does today.

### Layer 1 — the dmabuf probe uses implicit modifiers (fixable here)
`run_gl_test()` in `src/linux_util/src/dmabuf_probe.rs` allocates its test buffer with `gbm_bo_create()` (implicit modifier) and imports it via EGL with no `EGL_DMA_BUF_PLANE0_MODIFIER_*_EXT`. NVIDIA's EGL/GBM stack rejects implicit-modifier import, so `glEGLImageTargetTexture2DOES` fails with `GL_INVALID_OPERATION` (0x502) and `shared_texture_supported` latches off. Mesa accepts implicit modifiers, so this only bites NVIDIA — matching #157, closed as a side effect of #423 rather than actually fixed. The Vulkan import path in `gpu_paint/src/dmabuf_import.rs` already negotiates explicit modifiers; the EGL probe never caught up.

### Layer 2 — CEF OSR shared textures are broken on NVIDIA upstream (not fixable here)
[...]

Issue #157, opened 2026-04-13
Falls back to software rendering for CEF


lkcv:
**Describe the bug**
In both wayland and xwayland modes, CEF complains about shared textures not being supported and falls back to software rendering. This makes the UI very choppy.

mpv playback itself in the client seems fine.

**Steps to reproduce**
1. Run the appimage from the terminal with or with any ozone flags
2.  Verify in the logs that CEF is using software rendering

**System information**
 - Installation Method: AppImage
 - OS: Arch with Hyprland
 - GPU: NVIDIA RTX 4080 with 595.58.03 drivers
 - (Linux Only) **Wayland**

**Additional context**
Startup logs

09:59:34.268 [Main] INFO: jellyfin-desktop 0.0.0-dev (89dd886)
09:59:34.268 [Main] INFO: CEF 146.0.10+g8219561+chromium-146.0.7680.179
09:59:34.268 [Main] INFO: Display backend: wayland
09:59:34.279 [Main] INFO: Waiting for mpv window...
09:59:35.190 [Platform] INFO: dmabuf probe: testing on Wayland EGL display
09:59:35.236 [Platform] INFO: dmabuf probe using render node: /dev/dri/renderD128
09:59:35.236 [Platform] WARN: dmabuf probe: glEGLImageTargetTexture2DOES failed (0x502)
09:59:35.236 [Platform] WARN: dmabuf probe: ARGB8888 dmabuf import failed
09:59:35.236 [Platform] INFO: Shared textures not supported; using software CEF rendering
09:59:35.236 [Platform] DEBUG: set_titlebar_color(10,10,10) palette=(nil)
09:59:35.237 [Main] INFO: Platform init ok
09:59:35.238 [Main] INFO: [FLOW] calling CefInitialize...
[66319:66319:0413/095935.758798:ERROR:base/memory/shared_memory_switch.cc:289] Failed global descriptor lookup: 7
09:59:35.800 [Main] INFO: [FLOW] CefInitialize returned ok
09:59:35.801 [CEF] INFO: [CLIENT] Client::resize logical=105x115 physical=210x230 browser=(nil)
09:59:35.801 [Main] INFO: [FLOW] CreateBrowser(main) url=http://localhost:8096 lw=105 lh=115 pw=210 ph=230
09:59:35.801 [Main] INFO: [FLOW] CreateBrowser(main) call returned
09:59:35.801 [Main] INFO: [FLOW] CreateBrowser(overlay)
[...]

PR #633, opened 2026-07-23
linux_util: negotiate explicit DRM modifiers in dmabuf probe + guard NVIDIA OSR

mysticalsoap:
## Problem
On NVIDIA/Wayland the dmabuf capability probe (`src/linux_util/src/dmabuf_probe.rs`) allocates its test buffer with an implicit modifier and imports it via EGL without negotiating an explicit one. NVIDIA rejects implicit-modifier import, so `glEGLImageTargetTexture2DOES` fails with `GL_INVALID_OPERATION` (0x502) and shared textures latch off (same signature as #157). The Vulkan import path in `gpu_paint/dmabuf_import.rs` already negotiates explicit modifiers; the EGL probe didn't.

Fixing that alone regresses NVIDIA, though: once the probe passes, CEF's OSR compositor can't wrap the buffer as a writable render target ("Unable to initialize SkSurface" in Chromium's `shared_image_representation.cc`), blanking the window. That's an upstream Chromium bug CEF and Electron both inherit — Electron closed the identical failure not-planned (electron/electron#49247). Full write-up and two-layer diagnosis in #632.

## Fix
Two commits:
1. Probe negotiates an explicit modifier: query `eglQueryDmaBufModifiersEXT` for ARGB8888 (non-external-only), allocate via `gbm_bo_create_with_modifiers`, pass it via `EGL_DMA_BUF_PLANE0_MODIFIER_{LO,HI}_EXT`. Missing symbols or no advertised modifier fall through to the existing implicit path, so Mesa is unchanged.
2. Query `EGL_VENDOR` and block shared textures on NVIDIA regardless of the probe result, so the corrected probe can't hand CEF the OSR path that blanks the window.

Net-neutral by design: NVIDIA already fell back to software before this (via the 0x502 failure) and still does; Mesa already worked and is untouched. The value is a correct, Vulkan-path-consistent probe plus a guard that's ready for whenever the upstream OSR bug is fixed. Happy to drop the code and keep just the issue if you'd prefer.

## Verification
`just fmt`/`just lint`/`just build` pass. On an RTX 4090 + KWin (Plasma 6.7.3, driver 610.43.03): commit 1 alone makes the probe negotiate `0x300000000606013` and pass, then blanks with repeated SkSurface errors; commit 2 keeps the probe passing but blocks shared textures (one warn log), zero SkSurface errors, UI renders on the software path as before.

## Status
Draft — this doesn't improve NVIDIA performance (blocked upstream), it replaces an unexplained transport failure with a diagnosed one and lands groundwork. Flagging the `EGL_VENDOR` string check as the main thing to sanity-check before this is ready.

Assisted-by: claude-sonnet-5

### DIFF
--- a/src/linux_util/src/dmabuf_probe.rs
+++ b/src/linux_util/src/dmabuf_probe.rs
@@ -8,13 +8,16 @@
 //! `--ozone-platform=x11` over XWayland).
 
 use crate::egl_dyn as egl;
-use libloading::Library;
+use libloading::{Library, Symbol};
 use std::ffi::{CStr, CString, c_char, c_int, c_uint, c_void};
 use std::os::fd::RawFd;
 use std::ptr;
 
 // ARGB8888 fourcc — pulled from drm_fourcc.h to avoid a libdrm dep.
 const DRM_FORMAT_ARGB8888: u32 = 0x3432_5241;
+// Sentinel meaning "no explicit modifier" — never a real advertised modifier,
+// filtered out of eglQueryDmaBufModifiersEXT results defensively.
+const DRM_FORMAT_MOD_INVALID: u64 = 0x00ff_ffff_ffff_ffff;
 const GL_TEXTURE_2D: c_uint = 0x0DE1;
 const GL_NO_ERROR: c_uint = 0;
 const GBM_BO_USE_RENDERING: u32 = 0x0002;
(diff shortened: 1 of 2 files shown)


mysticalsoap:
Regarding the NVIDIA guard in this PR: the bug it works around is being fixed upstream (chromiumembedded/cef#4238, backed by https://chromium-review.googlesource.com/c/chromium/src/+/8220427).

Guard portion will just need to be removed once we get the CEF change here.

PR #655, opened 2026-08-02
feat(discord): publish the playing title, episode and position to Discord

Zaphkiel-Ivanovna:
Adds Discord Rich Presence: series and episode, or film and year, plus the playback position. New `jfn-discord` crate, no new external dependency, off by default.

image

## Action required from the maintainer

Inert until a Discord application exists. `DEFAULT_APPLICATION_ID` is left empty so nothing ships pointing at a personal account.

1. Create an app at <https://discord.com/developers/applications>. Its **name** is what Discord renders as "Watching *name*", so name it `Jellium Desktop`.
2. Upload `resources/discord/logo.png` and `pause.png` under **Rich Presence → Art Assets**. Asset names must be `logo` and `pause`; Discord takes them from the filenames, so upload as-is.
3. Set the Application ID in `DEFAULT_APPLICATION_ID` (`src/discord/src/sink.rs:14`).

The ID is public in every presence payload, not a secret. Merging before step 3 is harmless.

## Implementation

- Registers through `jfn_playback::register_event_sink`, not `MediaSink`: Discord is cross-platform, that trait gives one sink per platform and is already taken.
- Mirrors `jfn-mpris`: pure `projection` module for the rules, dedicated thread for the socket.
- IPC in tree (length-prefixed JSON over Unix socket or named pipe) rather than a dependency in a workspace with no async runtime.
- Reads on their own thread: Discord echoes every command and expects `PONG`, so an undrained socket blocks writes, and `stop()` must not wait on I/O. Backoff is loop state, not a sleep.
- Absolute timestamps, so nothing is pushed on a timer. The bar is omitted unless playback is steadily rolling, the only way to stop the client extrapolating.
- `MediaMetadata` gains seven fields `parse_metadata_json` already received and dropped. `Input::Metadata` boxed for `clippy::large_enum_variant`.
- Flatpak gains `--filesystem=xdg-run/discord-ipc-0`; the socket is invisible from the sandbox otherwise.

## Settings

New **Discord** section in Client Settings, mirrored in `settings.json`. Read at startup, so toggling needs a restart.

| Key | Type | Default |
| --- | --- | --- |
| `discordRichPresence` | bool | `false` |
| `discordApplicationId` | string | empty, falls back to the built-in ID |

## Your call

The large image is the Jellyfin poster URL. Discord's proxy embeds the origin host, so the **server hostname becomes visible** in the payload, and it only resolves for a publicly reachable server. One branch in `projection.rs` makes it logo-only.

## Testing

- 51 unit tests in `jfn-discord`, 272 workspace-wide, `just lint` clean.
- macOS against a live client: play, pause, seek, speed change, track change, Discord restarted mid-playback.
- Linux and Windows compile and pass tests but are untested at runtime; the Windows named pipe has never run against a real client.


andrewrabert:
This isn't something I want to add into the codebase proper due to reliance upon a third party API. 

Please open an issue for Discord support and I'll mark it as blocked by https://github.com/andrewrabert/jellium-desktop/issues/371

PR #657, opened 2026-08-02
Guard playback transparency against custom theme CSS

inkly:
Server-side themes/plugins can set an opaque background on body with !important, which paints over the transparent document during playback and hides the mpv video behind the web layer (black video, audio fine). Inject a small higher-specificity stylesheet scoped to transparentDocument so the video always shows through.


andrewrabert:
Example theme/plugin?

PR #674, opened 2026-08-19
Implement native-backed server switching

jacobleft:
## Summary
- Advertise Jellyfin Web `multiserver` support again, but intercept `selectserver` / `addserver` route changes before Jellyfin Web runs its in-page cross-origin probes.
- Add a web IPC (`showServerOverlay`) that opens Jellium's existing native server-selection overlay from the loaded Jellyfin Web UI.
- Add startup vs. manual-switch overlay modes: startup still auto-connects the saved server, while manual switching pre-fills the current server, supports cancel without resetting the main browser, and only saves/navigates after the native probe succeeds.
- Add a web shim regression test and `just test-web` recipe for the advertised feature + route-guard behavior.

## Why
Jellyfin Web's own add/select server flow probes replacement servers from the currently loaded server origin. With normal CSP/CORS policies, switching from one server origin to another local/private origin can fail before Jellium gets a chance to reload the replacement server as the top-level origin.

This routes server switching through Jellium's native probe instead. The native side validates the entered server URL, persists the normalized base URL, and loads it in the main browser as the new top-level page.

## Verification
- `just test-web`
- `cargo fmt --manifest-path src/Cargo.toml --all -- --check`
- `cargo test --manifest-path src/Cargo.toml -p jfn-cef`
- `cargo test --manifest-path src/Cargo.toml --workspace --no-fail-fast`
- `cargo clippy --manifest-path src/Cargo.toml --workspace --all-targets -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic`
- `cargo xtask fetch-cef && cargo xtask install --mpv-cli --prefix build/output`
- macOS bundle smoke via `open -n build/output/Jellium Desktop.app --args ...`: CEF initialized, remote debugging opened, and the app created the main browser at `http://10.88.0.2:8096` plus the startup overlay. A full live switch smoke was not completed because `http://10.88.0.2:8096/System/Info/Public` timed out from this machine during the run.

### DIFF
--- a/justfile
+++ b/justfile
@@ -40,6 +40,11 @@ clean:
 test: build
     cargo test --manifest-path src/Cargo.toml --workspace
 
+# Run web shim tests
+[group('test')]
+test-web:
+    node src/web/native-shim.test.mjs
+
 # Run tests (loads MSVC + bindgen libclang env via dev/windows/env.ps1)
 [group('test')]
 [windows]
--- a/src/jfn_cef/src/app.rs
+++ b/src/jfn_cef/src/app.rs
@@ -291,7 +291,12 @@ wrap_render_process_handler! {
                 "savedServerUrl" => {
                     let Some(args) = args else { return 1 };
                     let url = userfree_to_string(&args.string(0));
-                    call_js_global_string(frame, "_onSavedServerUrl", &[Arg::Str(&url)]);
+                    let switch_server = args.bool(1) != 0;
+                    call_js_global_string(
+                        frame,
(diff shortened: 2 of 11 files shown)


andrewrabert:
Hey thanks for this, but I'm currently in the process of rewriting the client config/about/CSD/server selection in Rust+Iced https://github.com/andrewrabert/jellium-desktop/pull/669

I already have a few ideas planned for how the multi-server support will work, but thanks for reminding me of the `selectserver` and `addserver` bits :)

